PR #241, opened 2014-10-22
Destroy active connections when stopping the server

### LINKED ISSUE #239: API Server takes 2 minutes to close/shutdown after last request
I noticed that when trying to stop and restart the server it takes 2 minutes before the API server can be shutdown. It seems to be due to a connection that is still open for 2minutes. no new connections are accepted but existing ones can still be used. If an new request comes through the API server needs to wait another 2minutes before it is closed successfully. 

This means that when stopping the servers, the API server could be kept alive when request are submitted, while the other server have already stopped

rcrichton:
Closes #239 

I'm not 100% happy with this implementation. What I really wanted to do was `.end()` each active socket but when I try to do that I get a segmentation fault from one of nodes native modules. I can't figure out what is causing this so I've changed the code to `.destroy()` active sockets rather. I don't really like this as if data is currently being send along this socket then it will be lost or cut off. This could be ok if we give an appropriate warning but it would be best if we could request that the socket closes and let it finish up before shutting down.

I not sure what we should do in this case.


rcrichton:
Hey @BMartinos, one thing you may want to do is try change this to how I originally intended. I wanted to use the `.end()` method rather than the `.destroy()` here: https://github.com/jembi/openhim-core-js/pull/241/files#diff-af5bc8ffc766db934873618fe4c339deR251

Perhaps you can try change this and see if it works with the new node.

BMartinos:
@rcrichton can we close the PR as this has been resolved in PR https://github.com/jembi/openhim-core-js/pull/396?

rcrichton:
Yup, we can, closed.

PR #304, opened 2014-12-12
Npm package

### LINKED ISSUE #282: Update package.json so that we can release to npm

rcrichton:
Closes #282 

@devcritter can you please review this. Also, @BMartinos @debonair @armageddon this may be of interest.

To test this run:

`sudo npm install -g .` form the root directory of the source.

Then run `openhim-core` (from anywhere, it should be put in your path) and the server should startup, with forever :)

Also, to symlink you current working directory to the package run `npm link`, then any changes you make are instantly accessible via the binary. Also, forever is set to watch files so as soon as it sees newly compiled files it will restart the server!


package.json
```diff
@@ -1,14 +1,19 @@
 {
-  "name": "openhim-core-js",
+  "name": "openhim-core",
   "description": "The OpenHIM core application that provides logging and routing of http requests",
-  "version": "0.1.0",
+  "version": "1.1.0",
   "private": true,
```
rcrichton: Yeah that's right. I'l switch this once we are ready to do a npm release. This prevents us from being able to release by accident before we are ready.

rcrichton:
@devcritter I've fixed the need for forever globally, my mistake (just run an `npm install`). I don't get that config error that you get, please can you try again and check if you still get that.

I've made sure node arguments get passed through so the custom conf works now. I've also merged with the latest in master.

`openhim-core --conf=/path/to/config.json`

rcrichton:
Hmm, I see tests are failing since the merge. They work locally for me. Will see if I can figure this out.

rcrichton:
Ok, tests are passing again.

package.json
devcritter: Cool

devcritter:
@rcrichton ok cool, I don't get that error anymore. But afraid I ran into another issue...

When running `openhim-core` outside of the src directory, I get:
```
fs.js:455
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^
Error: ENOENT, no such file or directory 'tls/key.pem'
    at Error (native)
    at Object.fs.openSync (fs.js:455:18)
    at Object.fs.readFileSync (fs.js:313:15)
    at Object.<anonymous> (/usr/local/node/node-v0.11.14/lib/node_modules/openhim-core/lib/middleware/router.js:31:13)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
Forever detected script exited with code 1
```

bausmeier:
Sounds like the paths need to be prepended with __dirname.

rcrichton:
Ok, I've added proper support for paths by introducing a appRoot global.

devcritter:
:+1: Awesome, worked for me now!

Issue #315, opened 2015-01-05
Investigate TCP channel issues


devcritter:
See email below from Justin:
> I am running a seed script which fires off random ADT_A01 messages (random names, dob, etc.) in parallel (8 threads on two machines). When I run this against the CR directly everything works, however when I do this via a passthrough on the HIM it works for a few minutes and then stops. On the HIM’s log I see this:
 
> info: Routing mllp request
info: Opened mllp connection to 142.222.45.50:2100
info: Received response end character
info: Storing response for transaction: 54a6f0a13c545e324f4c5b68
info: sending counts to statsd for ohie-him.marc-hi.ca.Development.54a5993ca6bbc665396b8db4
info: sending durations to statsd for ohie-him.marc-hi.ca.Development.54a5993ca6bbc665396b8db4
info: Storing visualizer events for transaction: 54a6f0a13c545e324f4c5b68
events.js:85
      throw er; // Unhandled 'error' event
            ^
Error: read ECONNRESET
    at exports._errnoException (util.js:742:11)
    at TCP.onread (net.js:535:26)

It could be an issue with the TCP server sockets not closing properly.

devcritter:
This could be related to #272

bausmeier:
Closed in https://github.com/jembi/openhim-core-js/commit/cd34fd45b3ecf0b2457cae864f8d49707db3b62e

Issue #272, opened 2014-11-21
Unhandled error event on tcpSockets


rcrichton:
events.js:85
      throw er; // Unhandled 'error' event
            ^
Error: This socket has been ended by the other party
    at Socket.writeAfterFIN [as write] (net.js:274:12)
    at IncomingMessage.<anonymous> (/home/ryan/git/openhim-core-js/lib/tcpAdapter.js:152:19)
    at IncomingMessage.emit (events.js:129:20)
    at _stream_readable.js:907:16
    at process._tickDomainCallback (node.js:411:11)

PR #317, opened 2015-01-06
Handle tcp errors and don't write to closed sockets

bausmeier:
* Checked to see if the socket is writable before writing to it in case the client has disconnected. (#272)
* Added an error listener so that connection errors do not crash the HIM. (#315)


bausmeier:
@devcritter Please take a look.

devcritter:
:+1: Great, thanks!

Issue #319, opened 2015-01-07
Update nodemailer and fix breaking changes


bausmeier:
Nodemailer introduced breaking changes in 1.0.0. See the [upgrade warning](https://github.com/andris9/Nodemailer/tree/c35d258b1a28689ecc4a3247d1f1da653b5a501a#upgrade-warning) and [migration guide](http://www.andrisreinman.com/nodemailer-v1-0/#migrationguide) for more details.

Issue #318, opened 2015-01-07
Update koa and fix breaking changes


bausmeier:
Koa introduced a breaking change (https://github.com/koajs/koa/commit/c2322f2b3dba5cd477c9fca0ae8879750abb2c68) in 0.8.0 which requires some refactoring. All usages of .status=string need to be changed to .status=number. I am not sure if there are any more changes that will break things past 0.8.0.

PR #320, opened 2015-01-07
Update dependencies

bausmeier:
Remove unused dependencies and update dependencies to the latest version where possible.

Koa and Nodemailer introduced breaking changes in newer versions which will require some more effort to update so I have created #318 and #319 for this.


bausmeier:
@devcritter Please review.

I made sure that all the tests are still passing and I have gone through the console and tested that everything I can think of is still working. Not sure how else I can make sure that this hasn't broken anything.

devcritter:
:+1: Looks good.

If you haven't done so yet, then maybe you can run some transactions against core in order to check that everything is good. The best would be against preprod after the merge.

PR #321, opened 2015-01-07
Fix puppet scripts

bausmeier:


bausmeier:
@devcritter Happy if I merge this since you have already tested it?

devcritter:
Yup, thanks. Go for it :+1:

PR #323, opened 2015-01-12
Statsd model

debonair:
@rcrichton please review


src/middleware/stats.coffee
```diff
@@ -16,16 +16,34 @@ exports.incrementTransactionCount = (ctx, done) ->
...
+        sdc.increment domain + '.Channels.' + ctx.authorisedChannel._id + '.nonPrimaryRoutes.' + route.name + '.StatusCodes.' + route.response.status # Per route response status
+
+        if route.orchestrations?
+          for orchestration in route.orchestrations
+            orchestrationStatus = orchestration.response.status
+            sdc.increment domain + '.Channels.' + ctx.authorisedChannel._id + '.nonPrimaryRoutes.' + route.name + '.orchestrations.' + orchestration.name
+            sdc.increment domain + '.Channels.' + ctx.authorisedChannel._id + '.nonPrimaryRoutes.' + route.name + '.orchestrations.' + orchestration.name + '.StatusCodes.' + orchestrationStatus
+            sdc.increment domain + '.Channels.' + ctx.authorisedChannel._id + '.nonPrimaryRoutes.' + route.name + '.StatusCodes.' + route.response.status + '.orchestrations.' + orchestration.name
+            sdc.increment domain + '.Channels.' + ctx.authorisedChannel._id + '.nonPrimaryRoutes.' + route.name + '.StatusCodes.' + route.response.status + '.orchestrations.' + orchestration.name + '.StatusCodes.' + orchestrationStatus
```
rcrichton: Minor comment: I see that sometime you are using a capital first letter in the namespace and sometime you aren't. I think we should standardise this to make it look better. Maybe we could rather just use a lowercase first letter everywhere instead.

src/model/statsd.coffee
```diff
@@ -0,0 +1,173 @@
...
+
+# Retrives Global Status Metrics from the StatsD API
+
+exports.fetcGlobalStatusMetrics = `function *(allowedIds) {
+  var filtersObject = this.request.query;
+  var userRequesting = this.authenticated,
+      path = '',
+      results = {},
+      final = {};
+  var allowedIds = allowedIds.length > 0 ? allowedIds : yield metrics.getAllowedChannelIDs(userRequesting)
```
rcrichton: Do you need the `var` here as this variable is passed in?

test/integration/statsDTests.coffee
```diff
@@ -0,0 +1,108 @@
...
+      Statsd = {}
+      mock = {}
+
+      `before(function(done){
+        Statsd = require("../../lib/model/statsd")
+        Statsd.authenticated = auth.getAuthDetails();
+        Statsd.request = {}
+        Statsd.request.query = {}
+        done();
+      })`
```
rcrichton: Why aren't these function written in coffeescript? They should be if possible.

test/integration/statsDTests.coffee
```diff
@@ -0,0 +1,108 @@
...
+      `after(function(done){
+        done();
+      });`
+
+      `it("should fetch global load Time metrics", function (done) {
+        mock = sinon.mock(Statsd);
+        mock.expects('fetchData').once().withExactArgs("/render?target=transformNull(summarize(stats.timers." + domain + ".Channels.mean,'1hour'))&from=-1days&format=json")
+        co(function* () {
+          yield Statsd.retrieveAverageLoadTimePerHour();
+        });
```
rcrichton: Try make only escape the generator function as javascript. The rest should rather be in coffeescript.

test/integration/statsDTests.coffee
```diff
@@ -0,0 +1,108 @@
...
+      `it("should fetch channel status metrics ", function (done) {
+          mock = sinon.mock(Statsd);
+          mock.expects('fetchData').once().withExactArgs("/render?target=transformNull(summarize(stats.counters." + domain + ".Channels.jjhreujiwh.Statuses.Processing.count,'1week'))&from=-1weeks&format=json");
+          co(function* () {
+             yield Statsd.retrieveChannelMetrics('status','jjhreujiwh');
+          });
+          mock.verify();
+          mock.restore();
+          done();
+        })`
```
rcrichton: Please can you add some more tests to test that the stats being fetched are being converted into the correct format for each of these functions.

src/model/statsd.coffee
```diff
@@ -0,0 +1,173 @@
+http = require 'http'
```
rcrichton: I don't think this file should be stored in src/model/ it doesn't really have anything to do with the model. It's more of a set of functions that are called for API. Perhaps, this should rather be moved to src/api/.

rcrichton:
Hey @debonair, this looks good and I'm excited to see it working! I have a few comment above that I think need to be addressed. Also, it may be worth comparing the stats that you get from this with the old stats that we were generating and make sure that you are getting the same results.

Also, are the old metrics endpoints only needed cos the console hasn't been updated? Can they be removed after that?

src/middleware/stats.coffee
debonair: Ok noted.

src/model/statsd.coffee
debonair: Ok yea, was wondering where that file should go.. am moving it to the api folder.

debonair:
Yes the old metrics endpoints will be removed when the core has been updated.

src/koaApi.coffee
```diff
@@ -77,10 +79,27 @@ exports.setupApp = (done) ->
...
   app.use route.get '/metrics/status', metrics.getGlobalStatusMetrics
   app.use route.get '/metrics/:type/:channelId', metrics.getChannelMetrics
-  
+# -- End of Old Metrics Routes --
+
+# -- New metrics Routes --
+  app.use route.get '/stats', if statsd_instance.enabled then statsd.retrieveTransactionCountPerHour else metrics.getGlobalLoadTimeMetrics
+  app.use route.get '/stats/status', if statsd_instance.enabled then statsd.fetcGlobalStatusMetrics else metrics.getGlobalStatusMetrics
+  app.use route.get '/stats/:type/:channelId', if statsd_instance.enabled then statsd.retrieveChannelMetrics else metrics.getChannelMetrics
+  app.use route.get '/stats/load-time', if statsd_instance.enabled then statsd.retrieveAverageLoadTimePerHour else metrics.getGlobalLoadTimeMetrics
```
rcrichton: Why is `metrics.getGlobalLoadTimeMetrics` used in both `/stats` and in `/stats/load-time` is this correct?

src/koaApi.coffee
```diff
@@ -77,10 +79,27 @@ exports.setupApp = (done) ->
   app.use route.get '/visualizer/events/:receivedTime', visualizer.getLatestEvents
   app.use route.get '/visualizer/sync', visualizer.sync
 
+# -- Old metrics Routes for Backward Compatibility
   app.use route.get '/metrics', metrics.getGlobalLoadTimeMetrics
   app.use route.get '/metrics/status', metrics.getGlobalStatusMetrics
   app.use route.get '/metrics/:type/:channelId', metrics.getChannelMetrics
-  
+# -- End of Old Metrics Routes --
```
rcrichton: Could we not rather just replace the metrics endpoints above with with a switch between the old metrics and the new statsd functions as you have done below? They should produce the data in the same format right? Then you could do away with the `/stats/*` routes and just include some extra routes for the statsd specific routes. Any reason why you can't do that?

src/middleware/stats.coffee
```diff
@@ -13,38 +13,76 @@ exports.incrementTransactionCount = (ctx, done) ->
...
+          sdc.timing domain + '.channels.' + ctx.authorisedChannel._id + '.orchestrations.' + orchestration.name, orchestratrionDuration
+          sdc.timing domain + '.channels.' + ctx.authorisedChannel._id + '.orchestrations.' + orchestration.name + '.statusCodes.' + orchestrationStatus , orchestratrionDuration
+
 
   catch error
-    logger.error error
+    logger.error error, done
 
 exports.koaMiddleware = `function *statsMiddleware(next) {
     timer = new Date();
```
rcrichton: You need to add tests for the above to functions.

test/integration/statsDTests.coffee
```diff
@@ -0,0 +1,132 @@
...
+        ]
+        convertToRequiredFormatSpy = sinon.spy Statsd, 'convertToRequiredFormat'
+        data = convertToRequiredFormatSpy fetchDataObject, 'retrieveAverageLoadTimePerHour'
+        data2 = convertToRequiredFormatSpy fetchDataObject, 'retrieveChannelMetrics'
+        data[0].avgResp.should.be.exactly 83
+        data2[0].avgResp.should.be.exactly 200
+        convertToRequiredFormatSpy.restore()
+        done()
+
+
```
rcrichton: Here I still think we need more tests. You are testing that the correct `fetchData` call is made but there are no tests to check that when data is returned from the `fetchData` function and it is processed that it is returned in the format that you expect. Please could you add some tests to test this.

I'm sure with sinon that you can make a mock that returns certain predefined data, then you would just need to check that that data gets returned in the format that you expect.

rcrichton:
Hey @debonair a few more comments above.

Also, a side note, with commit messages try not to repeat the same commit message over and over, rather for each commit clearly describe what you changed in that commit only.

src/koaApi.coffee
debonair: Yes this is correct I separated the functionality in the statsd api the  metrics.getGlobalLoadTimeMetrics returns both averageLoadTime  and load in one call

debonair:
@rcrichton please review

src/koaApi.coffee
```diff
@@ -77,10 +79,27 @@ exports.setupApp = (done) ->
...
+  app.use route.get '/metrics/status', if statsd_instance.enabled then statsd.fetcGlobalStatusMetrics else metrics.getGlobalStatusMetrics
+  app.use route.get '/metrics/:type/:channelId', if statsd_instance.enabled then statsd.retrieveChannelMetrics else metrics.getChannelMetrics
+  app.use route.get '/metrics/load-time', if statsd_instance.enabled then statsd.retrieveAverageLoadTimePerHour else metrics.getGlobalLoadTimeMetrics
+
+# -- StatsD only metrics routes --
+  app.use route.get '/statsd', statsd.retrieveTransactionCountPerHour
+  app.use route.get '/statsd/status', statsd.fetcGlobalStatusMetrics
+  app.use route.get '/statsd/:type/:channelId', statsd.retrieveChannelMetrics
+  app.use route.get '/statsd/load-time', statsd.retrieveAverageLoadTimePerHour
+  app.use route.get '/statsd/transactions/:channelId/:status', statsd.transactionsPerChannelPerHour
```
rcrichton: It looks like the last one `transactionsPerChannelPerHour` is the only statd only route. Can we not remove lines 95-98 as they are covered in 89-92.

src/koaApi.coffee
```diff
@@ -77,10 +79,27 @@ exports.setupApp = (done) ->
...
   app.use route.get '/visualizer/sync', visualizer.sync
 
-  app.use route.get '/metrics', metrics.getGlobalLoadTimeMetrics
-  app.use route.get '/metrics/status', metrics.getGlobalStatusMetrics
-  app.use route.get '/metrics/:type/:channelId', metrics.getChannelMetrics
-  
+# -- Old metrics Routes Purposely commented out for Backward Compatibility
+  app.use route.get '/mongometrics', metrics.getGlobalLoadTimeMetrics
+  app.use route.get '/mongometrics/status', metrics.getGlobalStatusMetrics
+  app.use route.get '/mongometrics/:type/:channelId', metrics.getChannelMetrics
```
rcrichton: Don't include these cos the tests are failing rather update the tests. Why are the tests failing, surely the data you get back should be the same?

test/fakeTcpServer.coffee
```diff
@@ -0,0 +1,102 @@
...
+dgram = require("dgram")
+
+# Start the server and listen for messages.
+#
+FakeServer::start = (cb) ->
+  that = this
+  @_socket = dgram.createSocket("udp4")
+  @_socket.on "message", (msg, rinfo) ->
+
+#    console.warn("Server got: '" + msg.toString() + "'");
```
rcrichton: Please remove this commented out code.

test/unit/statsMiddlewareTest.coffee
```diff
@@ -0,0 +1,91 @@
...
+                s.expectMessage domain + '.channels.ckjhfjwedsnfdsf.statuses.Successful.orchestrations.Lab API:1|c', ->
+                  s.expectMessage domain + '.channels.ckjhfjwedsnfdsf.statuses.Successful.orchestrations.Lab API.statusCodes.200:1|c', done
+
+#  it "Should measure transaction duration", (done) ->
+#    this.timeout = 900000
+#    stats.timer = new Date()
+#    stats.measureTransactionDuration ctx, () ->
+#      s._packetsReceived.forEach (item, index) ->
+#        console.log s._packetsReceived[index].toString()
+#      s.expectMessage '', done
```
rcrichton: Why is this code commented out? I think its a good idea to ensure the duration are send correctly.

rcrichton:
Hey @debonair I've still got some comments, see above. We can chat about this if you like? Ideally, the statd code should work seamlessly with the original metrics code, so that the user can switch between he two depending on their needs.

I like the inclusion of the fake server. That is a good idea.

src/koaApi.coffee
debonair: These tests fail because they inject data directly into the mongo database without running actual transactions, therefor, the metrics are never sent to the statsd server.

rcrichton: Ok, but can't you make these still work by setting the statsd config to false? Then it will just use the  original metrics methods?

debonair: I am not sure how to do this locally. as far as I know that config can only be set globally for all tests. Unless I put overide routes where a user can specifically request metrics from mongo, or from statsd

rcrichton: Hey @debonair have a look here for an example of how to change the config in tests: https://github.com/jembi/openhim-core-js/blob/master/test/integration/e2eIntegrationTests.coffee#L24-L25

debonair:
@rcrichton please review.

rcrichton:
Thanks @debonair this looks good to me! :+1:

PR #328, opened 2015-01-16
Mllp tls

### LINKED ISSUE #327: MLLP routes should support TLS

rcrichton:
Closes #327 

@bausmeier perhaps you could review this as you have been doing some of TLS work.

### DIFF
--- a/src/middleware/router.coffee
+++ b/src/middleware/router.coffee
@@ -152,7 +152,7 @@ buildNonPrimarySendRequestPromise = (ctx, route, options, path) ->
     handleServerError ctx, reason
 
 sendRequest = (ctx, route, options) ->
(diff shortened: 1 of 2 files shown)


bausmeier:
@rcrichton I will take a look.

bausmeier:
Looks good to me. I would suggest adding a test for routing to MLLP routes. Otherwise :+1:

rcrichton:
Cool, yeah that's probably a good idea.

rcrichton:
@bausmeier I've added a unit test for MLLP routes. I don't think we need to test TLS as that is handled by the other TCP tests now that the code is the same.

bausmeier:
:+1:

PR #330, opened 2015-01-22
Channels deleted flag

### LINKED ISSUE #324: Dealing with delete channels
A point that @BMartinos raised: how should we deal with deleted channels? In particular channels that have existing transactions?

Currently when a channel is deleted, the existing transactions will point to a non-existent channel. This means we can't display the channel names in the console or rerun those transactions (which is probably right). In addition these transactions won't show up on the dashboard (mongo aggregation) metrics.

So
1. Is this acceptable and we should leave it as is, or
2. should channels that have transactions linked against them be prevented from being deleted, or
3. have a flag that just marks them as deleted, or
4. should we store the channel name on a transaction in addition to the id, so that if deleted, we still have a reference?

devcritter:
Closes #324 

@rcrichton if you could please review?

Tested everything through the API and seems to be working fine.

Currently when someone tries to delete a channel, it will first check to see if there are any linked transactions. If not it will still properly remove the channel (useful if they added a bad channel or something), otherwise it sets the flag.

### DIFF
--- a/src/middleware/authorisation.coffee
+++ b/src/middleware/authorisation.coffee
@@ -64,6 +64,11 @@ if process.env.NODE_ENV == "test"
   exports.matchJsonPath = matchJsonPath
   exports.extractContentType = extractContentType
 
+# Is the channel enabled?
+# If there is no status field then the channel IS enabled
+checkChannelStatus = (channel) -> not channel.status or channel.status is 'enabled'
+
+
 exports.authorise = (ctx, done) ->
   Channel.find {}, (err, channels) ->
     for channel in channels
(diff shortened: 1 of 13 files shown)


src/middleware/authorisation.coffee
```diff
@@ -85,8 +90,8 @@ exports.authorise = (ctx, done) ->
               # deny access to channel if the content type isn't set
               continue
 
-          # now check if message content matches
-          if matchContent(channel, ctx.body) is true
+          # now check if message content matches and that the status is 'enabled'
+          if matchContent(channel, ctx.body) and isChannelEnabled channel
```
rcrichton: Perhaps we should check if a channel is enabled first rather than doing all this (possibly expensive) processing only to find out a channel is disabled?

rcrichton:
Great work, looks awesome. Just one comment above, otherwise :+1:

src/middleware/authorisation.coffee
devcritter: Yeah, good idea

PR #331, opened 2015-01-22
OHM-23: Vagrant -> added more strict time sync

### LINKED ISSUE #329: Vagrant: Time issues
When firing up a new vagrant instance, its time is a little behind the actual current time. Most likely this would mean that the instance hasn't got NTP sync turned on. The time should be up to date, since it causes occasional login problems from console.

devcritter:
Closes #329 

@rcrichton if you could please take a quick glance? Maybe @psbrandt too if you have a moment?

A bit hacky, but not too sure if there's a more elegant way...


rcrichton:
Looks alright to me :+1:

PR #332, opened 2015-01-23
Added upstart config

### LINKED ISSUE #303: Create a upstart config that uses the openhim binary created by npm install

rcrichton:
Closes #303 

@devcritter please can you review this?

### DIFF
--- a/resources/openhim-core.conf
+++ b/resources/openhim-core.conf
@@ -6,5 +6,5 @@ start on runlevel [2345]
 stop on runlevel [!2345]
 
 script
-  NODE_ENV=production openhim-core 2>&1 >> /var/log/openhim-core.log
+  NODE_ENV=production openhim-core >> /var/log/openhim-core.log 2>&1
 end script
--- a/resources/openhim-core.conf
+++ b/resources/openhim-core.conf
@@ -1,0 +1,10 @@
+# OpenHIM server upstart config
+
+description "OpenHIM server"
+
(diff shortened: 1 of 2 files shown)


resources/openhim-core.conf
```diff
@@ -0,0 +1,10 @@
+# OpenHIM server upstart config
+
+description "OpenHIM server"
+
+start on runlevel [2345]
+stop on runlevel [!2345]
+
+script
+  NODE_ENV=production openhim-core 2>&1 >> /var/log/openhim-core.log
```
devcritter: I think the 2>&1 should come at the end, i.e.
```
NODE_ENV=production openhim-core >> /var/log/openhim-core.log 2>&1
```

devcritter:
Very cool! Just a comment above, otherwise it looks good :+1: 

Just to note that I had an issue in vagrant, but that's due to it not finding the correct node path due to the way puppet sets it up. It's similar to the issues we had with the deployment puppet scripts, so I think we can consider it a puppet issue.

resources/openhim-core.conf
rcrichton: Ah, I think you are right, I'l fix it.

litlfred:
Very cool!

rcrichton:
Yeah, with general node install it should work fine. Perhaps, we should think about a way to specify an alternate path if this turns out to be an issue others face.

@litlfred glad you like it :)

PR #333, opened 2015-01-26
Fetch the channel from the db with each request

### LINKED ISSUE #326: Routes still send to an old endpoint when a channel's route has been updated
I saw this occur with a MLLP and previously with a TCP route. It may only affect those types.

To recreate:
* setup a channel with a MLLP route
* send a transaction through the channel
* update the channel's route with a new host and port
* send a transaction through the channel
* notice that it still sends to the old endpoint in the logs
* restart the server
* send a transaction through the channel
* notice that the endpoint that the transaction is being sent to is now correct

bausmeier:
Fixes #326


bausmeier:
@devcritter Please can you take a look at this. I have tested the scenarios in https://github.com/jembi/openhim-core-js/issues/326 and https://github.com/jembi/openhim-console/issues/208 and they are now working as expected.

devcritter:
:+1: Thanks, looks good to me!

PR #337, opened 2015-02-02
Tcp rerun

debonair:


debonair:
@rcrichton please review

rcrichton:
Hey @devcritter could you take this CR over from me? I have a few to many on my plate atm.

devcritter:
Sure, will take a look

devcritter:
Thanks @debonair it's looking good!

We just need to update the transaction model with the parentId and childIds for TCP/TLS transactions: https://github.com/jembi/openhim-core-js/blob/master/src/model/transactions.coffee#L37-L38 This'll allow for the console to display the original transactions.

debonair:
@devcritter I don't quite understand what you mean here.. I don't see where the HTTP transactions do these updates, or is this specific to TCP/TLS transactions?

devcritter:
@debonair so it's a common property for either http or tcp transactions. The http transactions are doing the updates in the middleware: https://github.com/jembi/openhim-core-js/blob/master/src/middleware/rerunUpdateTransactionTask.coffee

I think it's just in a different place since we're using special middleware for the http reruns.

debonair:
Right. I see. let me put that functionality in the worker somewhere.

debonair:
@devcritter please review, I have added the functionality you wanted.

devcritter:
:+1: Thanks, looks good!

PR #339, opened 2015-02-02
Export import

BMartinos:
This PR relates to issue: https://github.com/jembi/openhim-console/issues/210.

It was mentioned in the Jira ticket "OHM-24" that it will be easier for now to do this via mongodump and mongorestore


BMartinos:
Hey @devcritter Do you mind just reviewing this for me when you get a chance. Thanks

README.md
```diff
@@ -112,6 +112,24 @@ Then run start the server with:
 
 It will automatically startup on reboot.
 
+Exporting/Importing Default collections
```
devcritter: Maybe rather say something like "Exporting/Importing Server Configuration"

README.md
```diff
@@ -112,6 +112,24 @@ Then run start the server with:
...
+===========================
+
+### Exporting
+Follow the below steps to successfully export and import the default collections ( Users, Channels, Clients, ContactGroups, Mediators )
```
devcritter: Rather than saying default collections, let's clarify the task a bit: maybe say something like "Follow the below steps to export and import the server metadata configuration ... By default the Users, Channels, Clients, ContactGroups and Mediators collections will be exported"

resources/default-dump.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
devcritter: default-dump.sh feels a bit too generic to me. How about `openhim-configuration-export.sh`

devcritter:
Thanks @BMartinos, it's looking good! Just some comments.

BMartinos:
Hey @devcritter I have updated the readme as mentioned in the comments. Can you review again please. Thanks

rcrichton:
Hey @BMartinos, I'm sorry but having this as a script wasn't the intention of the issue. This was an issue in the console. The idea was to rather provide a user interface to enable an end user to easily import and export their channels. I think this PR is still useful but it doesn't solve the original issue.

BMartinos:
Hey @rcrichton Sure no problem. I will start working on creating this for the console. So the way im thinking of doing this is having a page for export/import. Export -> user selects options to export ( currently on 5 - channels, clients, users, mediators, contactGroups ). An API call is made to retrieve the collection and a export file with the JSON object is created for the user.

With the import, the user will upload the export file which will then either insert or update the existing collections. Do you have any other thoughts on this?

Should IDs be kept as in the original export or generated again when importing?

rcrichton:
Hi @BMartinos, yeah that sounds about right. I don't think we should think about importing or exporting mediators though. The mediators themselves should register themselves so we shouldn't need to import these. It would be useful for the user to be able to choose specifically which channels, user etc they wish to export but this can be left out in the first iteration if it will take too long in this sprint.

Also, I would suggest dropping all _id's from the export and have the import always add to the existing collection so items are never updated. This will make our lives easier I think.

Perhaps the export/import format could look like this:

```
{
  channels: [ ... ],
  users: [ ... ],
  clients: [ ... ],
  contactGroups: [ ... ]
}
```

devcritter:
@rcrichton @BMartinos just to mention that I think it will recreate the channels from the default mediator config if a mediator register itself again. I.e. after importing we'll end up with duplicate channels when the mediators register themselves again.

I do agree though that we should rather let mediators register themselves. I guess some options are:
1. Added handling to ensure that channels don't get recreated if they exist but the mediator doesn't
2. Export the mediators as well
3. Just leave it - users will manually have to delete duplicate channels

BMartinos:
I would go with option 2. I think this gives more flexibility/control to the user on what gets imported. Also, only an admin user will be allowed to do this import/export so they should be aware of a duplicate channel creation if they didn't export the mediators. ( channels and mediators aren't directly coupled together )

Not sure if this is a hidden bonus but even if a channel is duplicated by a mediator ( exactly the same ), transactions will only run against the first result it gets which will always be the same channel.

rcrichton:
Ok, @devcritter how does this work when a mediator registers itself on startup? If it starts multiple times does it create duplicates?

BMartinos:
@devcritter correct me if im wrong but I think what happens is when a mediator is started it send register details to openHIM. A check is done to see if the mediator exist and creates it if it doesn't ( based on URN ). If the mediator is restarted it will send register details again but will only update a mediator if the version has been bumped up. 

A channel only gets created when a mediator is created and not on an update.

devcritter:
Yes @BMartinos that's correct

litlfred:
Can we make the exclusion of _ids optionally (but default to exclude)?

Cheers,
-carl

On Feb 5, 2015 2:17 AM, Ryan Crichton <notifications@github.com> wrote:

Hi @BMartinos<https://github.com/BMartinos>, yeah that sounds about right. I don't think we should think about importing or exporting mediators though. The mediators themselves should register themselves so we shouldn't need to import these. It would be useful for the user to be able to choose specifically which channels, user etc they wish to export but this can be left out in the first iteration if it will take too long in this sprint.

Also, I would suggest dropping all _id's from the export and have the import always add to the existing collection so items are never updated. This will make our lives easier I think.

Perhaps the export/import format could look like this:

{
  channels: [ ... ],
  users: [ ... ],
  clients: [ ... ],
  contactGroups: [ ... ]
}

rcrichton:
Hey @litlfred, I'm sure we could do that. What use case do you envisage where we may want to keep these internal id's around?

BMartinos:
Hey @litlfred sure that can be added in. Would this be a "All or None" kind of option? or should it be collection specific? I would assume a use case could be to restore the HIM configuration to a previous date when an export was made, Updating the objects instead of inserting?

litlfred:
All or none should be fine.  And yes it would be for a restore.

Cheers,
-carl

On Feb 5, 2015 8:08 AM, Martin Brocker <notifications@github.com> wrote:

Hey @litlfred<https://github.com/litlfred> sure that can be added in. Would this be a "All or None" kind of option? or should it be collection specific? I would assume a use case could be to restore the HIM configuration to a previous date when an export was made, Updating the objects instead of inserting?

rcrichton:
Btw, I think this original PR can be merged with the mongodb export script. I still think it is useful. @devcritter is it a :+1:

devcritter:
:+1: definitely!

PR #359, opened 2015-02-12
OHM-27: Keystore API and TLS updates to use keystore

### LINKED ISSUE #299: Implement a 'keystore' were a user to can add certificates to authorise clients
We need to add a keystore to store certificates that are used to authorise clients. Here is a use case that I wrote (quickly, so mind the grammar...)

> I'm really thinking about this in the case of the IHE connectathon. I that case they give each client a cert. Each those certs are signed by their own connectatnon Certificate Authority (CA). So as the OpenHIM, I need to be able to say I accept all client that are signed by the CA (I won't know about each client's cert). We can do this by telling node we accept all client's that have certificates that are signed by the CA, this is done by importing the CA cert. Then for us to determine who the client is, we look at the certificate that they present, and match them on the CN of the cert and the domain of the client in our datastore.

rcrichton:
Closes #299 

@devcritter could you please review this for me? I'm doing this PR against concert-javascript as I wrote the generators in coffeescript so we didn't have to waste time converting them again.


src/api/keystore.coffee
```diff
@@ -0,0 +1,171 @@
...
+  ctx.status = status
+
+exports.getServerCert = ->
+  # Must be admin
+  if authorisation.inGroup('admin', this.authenticated) is false
+    logAndSetResponse this, 'forbidden', "User #{this.authenticated.email} is not an admin, API access to getServerCert denied.", 'info'
+    return
+
+  try
+    keystoreDoc = yield Keystore.findOne().exec()
```
devcritter: Very minor point: but maybe it would be good to add a projection for `cert`? That way it won't read out the server key as well.

Probably not a big deal ito performance, but that way the key won't be sitting in memory or get transferred in the db call. Probably also not a big deal ito security, but you never know.

src/middleware/router.coffee
```diff
@@ -76,53 +73,58 @@ sendRequestToRoutes = (ctx, routes, next) ->
...
-          # then set koa response from responseObj.response
-          setKoaResponse ctx, responseObj.response
-        else
-          setKoaResponse ctx, response
-      .fail (reason) ->
-        # on failure
-        handleServerError ctx, reason
-    else
-      promise = buildNonPrimarySendRequestPromise ctx, route, options, path
+  Keystore.findOne {}, (err, keystore) ->
```
devcritter: Maybe first do a check to see if there are any routes with `.cert` before doing the db call?

devcritter:
Looks good to me! Just some super-minor comments, but you can use your own discretion regarding those. Otherwise :+1:

src/middleware/router.coffee
rcrichton: Good idea, considering the benchmark performance.

src/api/keystore.coffee
rcrichton: Good idea, will do.

src/middleware/router.coffee
rcrichton: Actually I can't do this as I need the key and cert for the options of each request (https://github.com/jembi/openhim-core-js/pull/359/files#diff-ffc48656de2a6c6b5ee8a785b33109b0R88). I don't want to make this global as then if someone changes this in the UI it won't have immediate effect.

rcrichton:
Hey @devcritter, I've made some of the changes suggested about. Let me know if this is ok.

src/middleware/router.coffee
devcritter: aah right, sorry I missed that

devcritter:
Oh, I don't see any changes. Did you push anything?

rcrichton:
Oops, sorry forgot to push.

devcritter:
:+1: Looks great!

Issue #350, opened 2015-02-03
Convert all files in middleware/. to cofeescript (med)


rcrichton:
See epic here: https://github.com/jembi/openhim-core-js/issues/200

rcrichton:
Closed via b1a44f1aebed13c4d089d1e85d758940bdb90a4e

PR #362, opened 2015-02-12
OHM-54: Middleware convert js

rcrichton:
Closes: #350 

@devcritter another one for review.


src/middleware/visualizer.coffee
```diff
@@ -85,9 +85,7 @@ storeVisualizerEvents = (ctx, done) ->
...
 
-exports.koaMiddleware =  `function *visualizerMiddleware(next) {
-  yield next;
-  if (config.visualizer.enableVisualizer) {
-    storeVisualizerEvents(this, function(){});
-  }
-}`
+exports.koaMiddleware = (next) ->
+  yield next
+  if config.visualizer.enableVisualizer?
```
devcritter: I think you may need to take the question mark out, I suspect that'll just translate to `if (config.visualizer.enableVisualizer != null)`

src/middleware/tcpBypassAuthentication.coffee
```diff
@@ -16,15 +16,11 @@ exports.authenticateUser = (ctx, done) ->
...
+  yield authenticateUser this
 
-  if (this.authenticated) {
-    yield next;
-  } else {
-    this.response.status = "unauthorized";
-  }
-  
-}`
+  if this.authenticated?
```
devcritter: Same as the below comment regarding the question mark

devcritter:
Looks good to me! So much neater in coffeescript :) There's just a few places where you have `if variable?` but I believe that won't get interpreted as a boolean check, just nullity. Otherwise :+1:

rcrichton:
Yeah it really is! Good catch on the visualiser boolean. The others are all `this.authenticated` which is actually an client object or null/undefined if not authenticated, so I think the existential operator is ok there.

devcritter:
Aah, right. Cool, yeah

PR #369, opened 2015-02-13
Updated to stable node

devcritter:
@rcrichton bumped the puppet scripts to stable


rcrichton:
:+1:

debonair:
We should probably update the README and docs to stable as well

devcritter:
@debonair yeah definitely

Issue #325, opened 2015-01-15
Test if openhim runs on IO.js 1.0


debonair:
Test to see if the OpenHIM supports the recently released https://iojs.org/?1.0.0

debonair:
This has been tested and no issues were found.

PR #381, opened 2015-02-13
Add travis build target for node.js 0.12 and io.js

bausmeier:
Also cleaned up the prepublish and test scripts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "supertest": "^0.15.0"
   },
   "scripts": {
-    "prepublish": "./node_modules/grunt-cli/bin/grunt build",
-    "test": "./node_modules/grunt-cli/bin/grunt test"
+    "prepublish": "grunt build",
+    "test": "grunt test"
   }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: node_js
 node_js:
   - "0.11"
+  - "0.12"
+  - iojs
 services: mongodb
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -60,7 +60,7 @@
     "supertest": "^0.15.0"
   },
   "scripts": {
-    "prepublish": "./node_modules/grunt-cli/bin/grunt build",
-    "test": "./node_modules/grunt-cli/bin/grunt test"
```
rcrichton: Does this automatically run against the local grunt in node_modules? Or does it require grunt to be installed globally?

bausmeier:
@rcrichton We obviously want to support node 0.12, but I'm not sure about io.js. @debonair tested it out (https://github.com/jembi/openhim-core-js/issues/325) and didn't find any issues. Should we run the tests against io.js on travis or should I take it out?

rcrichton:
@bausmeier I think its fine to test against it. It may be something for us to consider running it on. I wonder if we should remove 0.11 as its unstable and rather just focus on 0.12 now.

package.json
bausmeier: Yes it adds dependencies to the path. See https://docs.npmjs.com/misc/scripts#path.

rcrichton: Oh cool. I didn't know it did that :)

bausmeier:
I think we should leave 0.11 in at least until we update the installation instructions in the readme.

rcrichton:
Cool, I'm fine with that.

PR #384, opened 2015-02-17
Public channels

### LINKED ISSUE #146: Support 'public' channels where request are not authenticated and authorised

debonair:
closes #146 

This is an initial version of my public channel solution. @rcrichton @devcritter @bausmeier @BMartinos what do you think


src/middleware/basicAuthentication.coffee
```diff
@@ -51,6 +51,7 @@ exports.authenticateUser = (ctx, done) ->
         done null, null
   else
     logger.error "No basic auth details supplied"
+    ctx.authenticated = {} # Set to empty object rather than null
```
rcrichton: I think null may work better.

rcrichton:
Hey @debonair, it look good so far. I'm just a bit unsure about the need for the whitelist. We already have 2 other authentication mechanisms, I'm unsure about adding a 3rd. Could you explain your thinking around this bit?

debonair:
The 3rd mechanism is for convenience really, if you want to configure the HIM want to whitelist transactions from within a local network perhaps without having to configure the clients that are calling the service with a username and password or certificate, because you already trust them as opposed to to having the channel fully public, I thought it would be a cool feature to have,

src/middleware/basicAuthentication.coffee
debonair: If I set it to null here, the middleware will not yield next

rcrichton:
@devcritter what do you think about this feature? I'm just worried about adding additional complexity and options for the user. But, I'm unsure, this does sound useful.

src/middleware/basicAuthentication.coffee
rcrichton: Yeah, but I think you should change the middleware then. Basically the middleware should always call yield next but only sometimes is a authenticated user set. As it is now, it will never get into the else statement of the middleware and that else block can probably be removed.

debonair:
@devcritter please review

src/middleware/authorisation.coffee
```diff
@@ -75,10 +75,16 @@ exports.authorise = (ctx, done) ->
...
       if pat.test ctx.request.url
-        matchedRoles = channel.allow.filter (element) ->
-          return (ctx.authenticated.roles.indexOf element) isnt -1
+        matchedRoles = {}
+        if ctx.authenticated?
+          if ctx.authenticated.roles?
+            matchedRoles = channel.allow.filter (element) ->
+              return (ctx.authenticated.roles.indexOf element) isnt -1
+        else
+          ctx.authenticated = {}
```
devcritter: Shouldn't this be set to null as per Ryan's comments? Or am I misunderstanding?

devcritter:
Thanks @debonair it's looking good.

A comment above and also: I tested on my system, and it worked perfectly if I had basic auth enabled, but if I just had mutualTLS enabled (i.e. basicAuth = false), the public channel transaction got rejected as unauthorized, even on http:5001 (I assume the https would get rejected in that case seeing as it's mutual auth)

rcrichton:
We should probably also default to having both basicAuth and TLSAuth enabled at the same time.

debonair:
@devcritter I have fixed the issues now.

src/middleware/authorisation.coffee
debonair: The messageStore middleware depends on having ctx.authenticated being set to something other than null, otherwise it throws an error, I have since fixed this in a later commit.

src/middleware/authorisation.coffee
```diff
@@ -75,10 +75,24 @@ exports.authorise = (ctx, done) ->
...
+          if ((channel.allow.indexOf ctx.authenticated.clientID) isnt -1)
+            allowedClient = true
+        else
+          # used by messageStore
+          ctx.authenticated =
+            ip: ctx.ip
+
         # if the user has a role that is allowed or their username is allowed specifically
-        if matchedRoles.length > 0 or (channel.allow.indexOf ctx.authenticated.clientID) isnt -1
+        if (matchedRoles.length > 0) or allowedClient or ((channel.whitelist.indexOf ctx.ip) isnt -1) or ((channel.authType == 'public') is true)
```
devcritter: I'm not too sure about the whitelist check here. Correct me if I'm wrong, but this means if the client ip is in the whitelist, they will _always_ be authenticated regardless of basic auth or tls certs?

Following our design call around this, my impression was instead that the whitelist is an _additional_ check, so if for example the channel is private, then the client needs to be authenticated via say basic auth _and_ their IP should be in the whitelist.

Maybe we need a check like? (code not tested!)
```
requireWhitelistCheck = channel.whitelist.length > 0
isWhiteListed = not requireWhitelistCheck or ((channel.whitelist.indexOf ctx.ip) isnt -1)
if isWhiteListed and ((matchedRoles.length > 0) or allowedClient or ((channel.authType == 'public') is true))
  ...
```

devcritter:
Thanks @debonair! Just a comment about the whitelist check

debonair:
@devcritter please take a look at this, I have implemented the code in your comment

devcritter:
:+1: Looks good to me!

PR #388, opened 2015-02-17
Switched to js bcrypt module

### LINKED ISSUE #387: Make bcrypt module optional (or even drop)?
The bcrypt module is pretty heavy dependency. It's a native module so you always need a C++ compiler on your system (please correct me if I'm wrong! but that's what it [looks](https://github.com/ncb000gt/node.bcrypt.js/tree/master/src) like to me). In Windows in particular it's pretty rough since you need  something like Visual Studio, making it very difficult to install the HIM in a windows server environment.

We already dropped support for bcrypt in the Console (https://github.com/jembi/openhim-console/commit/38f053922cb8bfa20b336c8b1590e5c4d9a0b5bf), so it's already an API level feature only anyway.

Therefore should we:
1. make bcrypt optional having a runtime check like http://stackoverflow.com/questions/15302618/node-js-check-if-module-is-installed-without-actually-requiring-it (which may be a bit ugly, since an implementer would need to install it manually)
2. remove support for it completely

I vote for 2.

devcritter:
Closes #387 

@rcrichton a quick one...

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "agenda": "^0.6.26",
     "basic-auth": "^1.0.0",
-    "bcrypt": "^0.8.0",
+    "bcryptjs": "^2.0.0",
     "co": "^4.1.0",
     "cookie": "^0.1.2",
     "forever-monitor": "^1.5.2",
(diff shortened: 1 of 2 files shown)


rcrichton:
@devcritter cool looks good :+1:

PR #400, opened 2015-02-22
Debian packaging

litlfred:
Seems to be working well on a clean Ubuntu 14.04 machine.


packaging/targets/trusty/etc/init/openhim-core.conf
```diff
@@ -0,0 +1,23 @@
...
+
+setuid openhim
+setgid openhim
+
+
+
+
+script
+  chdir /home/openhim/openhim-core-js
+  exec /home/openhim/.nvm/current/bin/node --harmony /home/openhim/openhim-core-js/lib/server.js  --conf=/etc/openhim/config.json
```
rcrichton: Here it may be better to use NPM to install the package then use the existing .conf file that we have in the repo.

packaging/targets/trusty/etc/openhim/config.json
```diff
@@ -0,0 +1,74 @@
...
+    "host": "127.0.0.1"
+  },
+  "application": {
+    "name": "Development"
+  },
+  "newUserExpiry": {
+    "duration": 7,
+    "durationType": "days"
+  }
+}
```
rcrichton: Is it not possible to copy the default.json file from the repo rather than maintain a copy?

rcrichton: Is it not possible to copy the default.json file from the repo rather than maintain a copy?

packaging/create-deb.sh
```diff
@@ -0,0 +1,181 @@
...
+    done
+    cd $HOME
+    
+    $GIT  status
+
+    $GIT add .
+
+    echo "Incrementing version"
+    $GIT commit ./ -m "\"${COMMITMSG}\""
+    $GIT tag $TAG
```
rcrichton: We are already tagging releases for our npm releases, could we not make use of those rather?

rcrichton: From looking a this more deeply, I'm not sure this is possible as we need extra files in packaging for the deb build and these need to be committed.

packaging/targets/trusty/debian/rules
```diff
@@ -0,0 +1,44 @@
...
+	dh_testdir
+	touch build-stamp
+
+clean: 
+	dh_testdir
+	dh_testroot
+	rm -f build-stamp
+	dh_clean 
+
+install: build
```
rcrichton: Could we make the install target just do a `npm install openhim-core -g`. This could simplify things greatly as we don't have to  deal with the source at all?

packaging/targets/trusty/debian/rules
```diff
@@ -0,0 +1,44 @@
...
+	dh_installdeb
+	dh_gencontrol
+	dh_md5sums
+	dh_builddeb
+
+# Build architecture-dependent files here.
+binary-arch:
+
+binary: binary-indep binary-arch
+.PHONY: build clean binary-indep binary-arch binary install
```
rcrichton: Are all the rest of these targets really needed. I don't quite understand them but I'm wondering if all of them are required for our simple install.

rcrichton:
Hey @litlfred I just have a few questions about this PR. I'd like to try simplify the deb if possible. Please let me know your thoughts.

litlfred:
Sure… what do you want to simplify?
Cheers,
-carl

rcrichton:
I've added a few comment in the code of the PR. Could you take a look and let me know if the things I've mentioned are possible. I'm still new to this stuff, so still trying to wrap my head around it fully.

packaging/targets/trusty/etc/init/openhim-core.conf
litlfred: That should be fine.  I think the only reason this was here was because I was installing under /home/openhim and needed a conf file to get the paths correct.

packaging/targets/trusty/etc/openhim/config.json
litlfred: Yes.  You can copy over the file by editing the targets/trusty/debian/postinst file

packaging/create-deb.sh
litlfred: I think keeping the tagging for Ubuntu packaging and npm packaging should be kept separate.  For another project, I tried to "reuse" the tags across different releases, but it ended up being more of a headache than it was worth.

The other main issue is that you may need to repackage for Ubuntu independent of the npm packaging.  For example if you want to keep the npm package as is, but repackage Ubuntu to add new functionality/fix bugs.

packaging/targets/trusty/debian/rules
litlfred: That would probably work.

packaging/targets/trusty/debian/rules
litlfred: I think some of this is used by the Launchpad PPA builder, but honestly.. not sure what is needed by what here.  I am not an expert in debian packaging.

litlfred:
I am newish to this stuff too :-)

litlfred:
Pull in recent changes

litlfred:
Pull in recent changes.

litlfred:
Oops.  didn't mean to close it.

PR #407, opened 2015-02-25
changed default config options 

debonair:
True for both auth methods and false for statsd @rcrichton review s'il vous plaît


rcrichton:
:) thanks @debonair :+1:

Issue #410, opened 2015-02-26
ChildIDs not being set for rerun transactions


devcritter:
The childIDs element is being set for rerunned transactions anymore.

ParentID is still being set correctly.

BMartinos:
So after a bit of investigation the problem is located at the updateTransaction function in the rerun middleware. https://github.com/jembi/openhim-core-js/blob/master/src/middleware/rerunUpdateTransactionTask.coffee#L8. 

The mongo validation seemed to be failing due to "request.method" or "request.path" doesnt exist

PR #412, opened 2015-02-26
Updated model required field for path and method

BMartinos:
This PR relates to issue: #410 

Some channels or orchestrations don't always have request.path or request.method value so the rerun update function didn't execute correctly and failed to update.


BMartinos:
Hey @rcrichton can you just review this for me. thanks

rcrichton:
:+1:

PR #414, opened 2015-02-27
Atna model

debonair:
@devcritter please review this atna message model


src/model/auditRecords.coffee
```diff
@@ -0,0 +1,56 @@
+mongoose = require "mongoose"
+Schema = mongoose.Schema
+
+eventIdSchema = new Schema
+  "code": { type: String, required: false }
+  "codeSystemName": { type: String, required: false }
+  "displayName": { type: String, required: false }
```
devcritter: Not a big deal, but we could probably use a single schema for all the types that use code, codeSystemName and displayName. A good name would be `CodedType` or something similar.

devcritter:
Thanks @debonair This looks really great!

I think it covers every field that at least I'm aware of. However I'm not familiar with the DICOM messages. Have you had a chance to look these yet to see if we'd be able to support both?

Also, have you started thinking about how we should store the raw message that we get? Should we maybe at a field here, or rather store it in another collection? I'm leaning toward the latter actually...

@rcrichton any thoughts from your side?

src/model/auditRecords.coffee
```diff
@@ -0,0 +1,52 @@
...
+
+ParticipantObjectIdentificationSchema = new Schema
+  "ParticipantObjectID": { type: String, required: false },
+  "ParticipantObjectTypeCode": { type: String, required: false },
+  "ParticipantObjectTypeCodeRole": { type: String, required: false }
+  "ParticipantObjectIDTypeCode": ParticipantObjectIDTypeCodeSchema
+  "ParticipantObjectQuery": { type: String, required: false }
+  "ParticipantObjectDetail": ParticipantObjectDetailSchema
+
+ParticipantObjectDetailSchema = new Schema
```
rcrichton: THis may need to move up above the 'ParticipantObjectIdentificationSchema'.

rcrichton:
Thanks @debonair looking good. We should be aiming to support the DICOM format but they are fairly similar, it may be good to check that we include all the DICOM fields as Hannes says. 

Also, it may be good to camel case all the field for consistency,

@devcritter about the `rawMessage` I think its fine to be just a string field in the schema. We can always project it out if we don't want to retrieve it. Also the messages aren't that big, I'd rather simplify things and have a string field than try maintain links to another collection. That said I don't think either way is a bad way of doing things. What do you think? Anything I'm missing?

devcritter:
@rcrichton nope sounds good. Those xds audit messages are quite big, but I don't think that changes anything. I think we could even project out by default.

BMartinos:
Hey @debonair @devcritter @rcrichton I made some updates to the Audit schema. Including it into the API broke the server. The schema logically looked correct but some small changes corrected the issue. New Schema may only be used on arrays/refs and not sub properties

rcrichton:
Cool, I suspected that may be the case but I wasn't sure.

devcritter:
Pushing this forward as #416 is blocked by this ticket.

@debonair @BMartinos an outstanding issue is camel-casing the field names as per Ryan's suggestion. The most important thing is consistency, see e.g. https://github.com/jembi/openhim-core-js/blob/master/src/model/channels.coffee. So I think the best is lower case for fields and capitals for schema names. I think it's fine to have something like `userID` even if the xml schema uses `UserID`.

@rcrichton thoughts? anything else outstanding from your side?

rcrichton:
@devcritter no this looks good to me. Just the camel casing issue. Otherwise it looks good.

BMartinos:
@rcrichton and @devcritter do you guys mind just reviewing this again. Thanks

BMartinos:
@devcritter  mind just reviewing this small change for me please. Thanks

devcritter:
:+1: Looks good to me!

litlfred:
Hey All.  Just wanted to make sure that the RESTful ATNA profile from IHE-ITI is on your radar as you do this:
   ftp://ftp.ihe.net/IT_Infrastructure/iheitiyr13-2015-2016/Technical_Cmte/Workitems/ATNA%20Repository-ATNA%20Query/

rcrichton:
Thanks @litlfred we do know about this. We are looking to support it once it matures a bit.

PR #416, opened 2015-03-03
Atna api repository

BMartinos:
This PR relates to issue: https://github.com/jembi/openhim-core-js/issues/376


BMartinos:
Hey @devcritter Do you mind just reviewing this PR for me when you get a chance. Thanks

devcritter:
Thanks @BMartinos. I'll look into more detail shortly, but I just wanted to raise the mongo db issue so long.

It might be a good idea for us to separate the audit data into another database. This is to keep it separate from the normal openhim data and make it more flexible in terms of management, ie. it gives you the option to keep the audit data on another server if you wanted to (a good practice pointed out to us by CSIR). The best would be to add a property in config with a different mongo string. This also gives users the option to easily point audit writes to another mongo node in the cluster.

@rcrichton @armageddon what do you think?

src/api/audits.coffee
```diff
@@ -0,0 +1,129 @@
...
+    when "simpledetails"
+      # view minimum required data for audit details view
+      return {}
+    when "full"
+      # view all audit data
+      return {}
+    else
+      # no filterRepresentation supplied - simple view
+      # view minimum required data for audits
+      return { "participantObjectIdentification": 0, "activeParticipant": 0, "rawMessage": 0 }
```
devcritter: A good field to exclude would be [participantObjectQuery](https://github.com/jembi/openhim-core-js/blob/master/src/model/audits.coffee#L24). That contains the actual message (pix, xds) that was sent. In fact we should probably project that out for all cases except `full`.

devcritter: Aah oops. You can ignore this :) I wasn't paying full attention to the schema, I see it's already excluded with participantObjectIdentification.

rcrichton:
@devcritter I think that seems like a good idea. I think the default should be to use the same db as the normal OpenHIM data but to allow for that as an option. Just to keep it as simple as possible for implementers getting started with the OpenHIM.

devcritter:
Cool, yeah good idea.

src/api/audits.coffee
```diff
@@ -0,0 +1,129 @@
...
+  
+
+
+###
+# Adds a Audit
+###
+exports.addAudit = ->
```
devcritter: Just to open this up to discussion - do you we want to support custom add endpoints for auditing (an addition to the standards-based ones)? I personally think it's okay to leave this in - we could view it as a "value add" service that the HIM is providing...

@rcrichton what do you think?

rcrichton: Yeah I agree.

devcritter:
@BMartinos looks good to me. No other comments for the moment other than addAudit discussion.

src/api/audits.coffee
BMartinos: Yip by default I just want to return the 'eventIdentification' details for the listings. I think that's all that would be needed to list the audits

BMartinos:
Hey @devcritter @rcrichton. Just want to pick this up again as ive have completed the first version of the ATNA UI and it uses this branch to call the API for audits. Is there anything that i have missed in the API that i still need to include? Like the separate database option?

devcritter:
Thanks, nothing from my side.

I started looking briefly at the database option: it seems pretty straightforward to have multiple db connections: http://blog.fluidapps.ca/handling-multiple-databases-and-connections-with-mongoose/
The implication though [of following the approach outlined in that guide] is that the existing `mongoose.model` calls would need to be replaced by a call to an openhim db connection call, i.e. now we'll have two connections: one for the main openhim db and another for the audit db. But should be straightforward - no objections from me if you want take this on.

rcrichton:
@BMartinos nothing more from my side. Just the multiple db option as Hannes suggests.

BMartinos:
Hey @devcritter Do you mind just reviewing this PR for me when you get a chance. I have included the separate mongo connection for ATNA records. I also found a bug with the rerun server not starting. the variable name was just a bit wrong. Thanks

src/api/audits.coffee
```diff
@@ -0,0 +1,169 @@
...
+    utils.logAndSetResponse this, 'forbidden', "User #{this.authenticated.email} is not an admin, API access to getAudits denied.", 'info'
+    return
+
+  try
+
+    # execute the query
+    eventID = yield Audit.distinct('eventIdentification.eventID').exec()
+    eventTypeCode = yield Audit.distinct('eventIdentification.eventTypeCode').exec()
+    roleIDCode = yield Audit.distinct('activeParticipant.roleIDCode').exec()
+    participantObjectIDTypeCode = yield Audit.distinct('participantObjectIdentification.participantObjectIDTypeCode').exec()
```
devcritter: Can we add `auditSourceId` to the filter options result as well? It would be really cool to have that as a dropdown on the frontend.

devcritter:
Thanks @BMartinos! Looks really good. I just added a comment about including the audit source id, which would be a really useful feature. Otherwise I think we're good to merge :+1:

devcritter:
:+1:

PR #420, opened 2015-03-05
Disabled make-install for node.js in puppet

devcritter:
@rcrichton just a quick one: disabled makeinstall for node in puppet. This improves vagrant startup times drastically


rcrichton:
:+1:

PR #421, opened 2015-03-05
Minor fix to transaction API - getTransactionById()

BMartinos:
Found this bug in transactions -> find by ID. 

The result returned is not an array but rather an object when searching for a specific record. If the result is null (not found) then a server error is thrown on .length. 

Replicate: View a valid transaction details. Change the URL to reference an invalid transaction ID and the labels are still show as well as an error in console.


BMartinos:
@devcritter do you mind just reviewing this quick PR for me. Thanks

devcritter:
:+1:

PR #423, opened 2015-03-06
Atna udp interface

### LINKED ISSUE #375: Create the SYSLOG ATNA interface

devcritter:
Closes #375 

### DIFF
--- a/src/auditing.coffee
+++ b/src/auditing.coffee
@@ -1,7 +1,10 @@
 logger = require 'winston'
+syslogParser = require('glossy').Parse
 Audit = require('./model/audits').Audit
 
-exports.processAudit = (msg) ->
+exports.processAudit = (msg, callback) ->
+  parsedMsg = syslogParser.parse(msg)
+
   audit = new Audit
     rawMessage: msg
 
@@ -10,3 +13,4 @@ exports.processAudit = (msg) ->
       logger.error "An error occurred while saving the audit entry: #{err}"
     else
       logger.info 'Processed audit message'
+    callback()
--- a/src/auditing.coffee
+++ b/src/auditing.coffee
@@ -1,0 +1,12 @@
+logger = require 'winston'
+Audit = require('./model/audits').Audit
+
+exports.processAudit = (msg) ->
+  audit = new Audit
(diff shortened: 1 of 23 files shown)


src/auditing.coffee
```diff
@@ -0,0 +1,72 @@
...
+    mergeAttrs: true,
+    explicitArray: false
+    tagNameProcessors: [firstCharLowerCase]
+    attrNameProcessors: [firstCharLowerCase, csdCodeToCode, originalTextToDisplayName]
+
+  parseString xml, options, (err, result) ->
+    return callback err if err
+
+    if not result?.auditMessage
+      return callback 'Document is not a valid AuditMessage'
```
rcrichton: Should we not create a new Error object here and set the message?

src/auditing.coffee
```diff
@@ -0,0 +1,72 @@
...
+  parseAuditRecordFromXML parsedMsg.message, (err, result) ->
+    audit = new Audit result
+
+    audit.rawMessage = msg
+    audit.syslog = parsedMsg
+    delete audit.syslog.originalMessage
+    delete audit.syslog.message
+
+    audit.save (saveErr) ->
+      if err or saveErr
```
rcrichton: Do we want to try save the audit if `err` exists?

rcrichton:
@devcritter looks great! Just some minor comments above, otherwise :+1:

src/auditing.coffee
devcritter: The use case I have in mind is, esp. with the UDP interface, if messages come through incomplete. That way we still store the raw message, even if we can't parse it. I saw this with a few times with OpenATNA and those big PnR messages. If we do get to the `err` stage, it also means it's valid syslog, so we won't be auditing garbage.

What do you think?

FYI looking at this code again, i see it's not too readable, just made a couple of small edits...

devcritter:
@rcrichton made some changes

rcrichton:
:+1:

src/auditing.coffee
rcrichton: I see. I didn't get that at first, but I think it's a great idea.

PR #424, opened 2015-03-09
Visualizer events: async processing. Replaced create with batch driver

devcritter:
insert.

@rcrichton if you could please take a look? I saw in the flame graph that the visualizer usage was quite high, not super high, but still. These few improvements shave off a few ms.

It seems mongoose's `create` just loops through the array and saves each document, rather than doing a proper batch insert. If switched this out with a driver batch insert. This mean we're not doing any validations, but seems okay to me since you can't add to this collection from the API, only from this code. To reflect our lack of validation, I took out the `require`s in the model.

### DIFF
--- a/src/middleware/visualizer.coffee
+++ b/src/middleware/visualizer.coffee
@@ -52,7 +52,7 @@ addRouteEvents = (dst, route, prefix, tsDiff) ->
   if endTS-startTS<minEvPeriod then endTS = startTS+minEvPeriod
 
   # Transaction start for route
-  dst.push new events.VisualizerEvent
+  dst.push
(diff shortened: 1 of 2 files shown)


rcrichton:
Cool looks good to me :+1:

devcritter:
Argh, sorry @rcrichton mind taking a quick look again?

I forgot to check that the TTL is working properly.

rcrichton:
Ok, looks good to me :+1: 

So does this just use the native driver though the mongoose model?

devcritter:
Yeah, it's a bit of a "under the hood" reference to the collection that mongoose is using.

Issue #403, opened 2015-02-25
Record metrics and visualiser data when updating transactions via the API


bausmeier:
This is not currently happening when updating the response and orchestration data of a transcation from a mediator.

PR #425, opened 2015-03-09
Updated transaction metrics

debonair:
Collects metrics for async transactions

### DIFF
--- a/src/api/transactions.coffee
+++ b/src/api/transactions.coffee
@@ -4,6 +4,14 @@ Q = require 'q'
 logger = require 'winston'
 authorisation = require './authorisation'
 utils = require "../utils"
+config = require '../config/config'
+statsd_client = require "statsd-client"
+statsd_server = config.get 'statsd'
+sdc = new statsd_client statsd_server
+application = config.get 'application'
+os = require "os"
+timer = new Date()
+domain = os.hostname() + '.' + application.name
 
 getChannelIDsArray = (channels) ->
   channelIDs = []
@@ -25,7 +33,7 @@ getProjectionObject = (filterRepresentation) ->
       # no filterRepresentation supplied - simple view
       # view minimum required data for transactions
       return { "request.body": 0, "request.headers": 0, "response.body": 0, "response.headers": 0, orchestrations: 0, routes: 0 }
-  
+
 
 
 
@@ -203,7 +211,7 @@ exports.findTransactionByClientId = (clientId) ->
       .find filtersObject, projectionFiltersObject
       .sort 'request.timestamp': -1
       .exec()
-    
+
   catch e
     utils.logAndSetResponse this, 'internal server error', "Could not get transaction by clientID via the API: #{e}", 'error'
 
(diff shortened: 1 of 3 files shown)


debonair:
@devcritter @rcrichton please review. We decided to  make the asyncronous response be a secondary route so as not to overwrite the original response. I am concerned though about sending mongo queries over the wire, because they require the mediator to know too much about the underlying data model @bausmeier  what do you think

src/api/transactions.coffee
```diff
@@ -220,12 +228,71 @@ exports.updateTransaction = (transactionId) ->
...
+    yield transactions.Transaction.findByIdAndUpdate transactionId, updates, () ->
+      that.body = "Transaction with ID: #{transactionId} successfully updated"
+      that.status = 'ok'
+      logger.info "User #{that.authenticated.email} updated transaction with id #{transactionId}"
+
+      ###
+      # Update transaction metrics
+      ###
+      transactions.Transaction.findById transactionId, (err, doc) ->
+        if updates['$push'].routes?
```
rcrichton: What is $push here. Is this a property someone has to set when updating or is it something different?

debonair: You have to set the push property so that you don't overwrite the other secondary routes with your update.

src/api/transactions.coffee
```diff
@@ -220,12 +228,71 @@ exports.updateTransaction = (transactionId) ->
...
   transactionId = unescape transactionId
   updates = this.request.body
+  that = this
 
   try
-    yield transactions.Transaction.findByIdAndUpdate(transactionId, updates).exec()
-    this.body = "Transaction with ID: #{transactionId} successfully updated"
-    this.status = 'ok'
-    logger.info "User #{this.authenticated.email} updated transaction with id #{transactionId}"
+    yield transactions.Transaction.findByIdAndUpdate transactionId, updates, () ->
```
rcrichton: Why did you need to change this to a callback instead of just having it as it was before. This should work the same way.

debonair: I needed the stats log actions to happen after the transaction has happened, i.e in that callback I made

src/api/transactions.coffee
rcrichton: Ah, I see so its a mongodb operator.

rcrichton:
Hey @debonair this looks good to me. You may want to try rather use string interpolation to make this look neater but that isn't required. Ie. `sdc.timing "#{domain}.channels.#{doc.channelID}.#{route.name}.orchestrations.#{orchestrationName}.#{metric.name}", metric.value`

Otherwise, :+1:

src/api/transactions.coffee
rcrichton: But the yield will wait until the promise is resolved before continuing. So I don't think you need the callback at all. It will do the same thing.

bausmeier: I am seeing some really weird behavior here with the yield and the callback. Errors updating the transaction aren't being handled and I am getting a combination of successful and failed responses and log messages.

src/api/transactions.coffee
```diff
@@ -220,12 +228,71 @@ exports.updateTransaction = (transactionId) ->
...
+      that.body = "Transaction with ID: #{transactionId} successfully updated"
+      that.status = 'ok'
+      logger.info "User #{that.authenticated.email} updated transaction with id #{transactionId}"
+
+      ###
+      # Update transaction metrics
+      ###
+      transactions.Transaction.findById transactionId, (err, doc) ->
```
bausmeier: You should handle the err argument.

debonair:
closes #403

PR #437, opened 2015-03-12
OHM-66: Added ATNA TLS/TCP interface with test - TLS Tests still failing

BMartinos:
This PR relates to issue: https://github.com/jembi/openhim-core-js/issues/374

### DIFF
--- a/src/server.coffee
+++ b/src/server.coffee
@@ -267,59 +267,18 @@ startAuditUDPServer = (auditUDPPort, bindAddress) ->
   return defer
 
 
-# function to start the TLS Audit server
-startAuditTlsServer = (auditTlsPort, bindAddress) ->
+
+# function to start the TCP/TLS Audit server
+startAuditTcpTlsServer = (type, auditPort, bindAddress) ->
   defer = Q.defer()
 
-  tlsAuthentication.getServerOptions true, (err, options) ->
-    return callback err if err
-
-    auditTlsServer = tls.createServer options, (sock) ->
-      message = ""
-      length = 0
-
-      sock.on 'data', (data) ->
-        # convert to string anf concatenate
-        message += data.toString()
-
-        # check if length is is still zero and first occurannce of space
-        if length == 0 and message.indexOf(' ') != -1
-          # get index of end of message length
-          lengthIndex = message.indexOf " "
-
-          # source message length
-          lengthValue = message.substr(0, lengthIndex)
-
-          # remove white spaces
-          length = parseInt(lengthValue.trim())
-          
-          # update message to remove length - add one extra character to remove the space
-          message = message.substr(lengthIndex + 1)
-        
-        # if sourced length equals message length then full message received
-        if length == message.length
-          logger.info "[Auditing TLS] Received message from #{sock.remoteAddress}"
-          auditing.processAudit message, -> logger.info "[Auditing TLS] Processed audit"
-
-      sock.on 'error', (err) ->
-        logger.error err
-    
-    auditTlsServer.listen auditTlsPort, '0.0.0.0', ->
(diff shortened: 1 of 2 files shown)


test/integration/auditingIntegrationTests.coffee
```diff
@@ -28,3 +40,65 @@ describe "Auditing Integration Tests", ->
...
+    it "should send TLS audit messages and save (valid)", (done) ->
+
+      options =
+        cert: fs.readFileSync "test/resources/trust-tls/cert1.pem"
+        key:  fs.readFileSync "test/resources/trust-tls/key1.pem"
+        ca: [ fs.readFileSync "test/resources/server-tls/cert.pem" ]
+
+      normalSocket = new (net.Socket)
+      client = new (tls.TLSSocket)(normalSocket)
+      client.connect 5051, 'localhost', options, ->
```
rcrichton: Sidenote: here I think you can just do `tls.connect ...` and delete the line above. Same on the TCP server.

src/server.coffee
```diff
@@ -262,6 +266,91 @@ startAuditUDPServer = (auditUDPPort, bindAddress) ->
 
   return defer
 
+
+# function to start the TLS Audit server
+startAuditTlsServer = (auditTlsPort, bindAddress) ->
```
rcrichton: I think you could combine the startAuditTlsServer and the startAuditTcpServer functions here. Rather just have a parameter for tls or not. There is a fair bit of code duplications between these two.

test/integration/auditingIntegrationTests.coffee
```diff
@@ -28,3 +40,65 @@ describe "Auditing Integration Tests", ->
...
+        messagePrependlength = testAuditMessage.length + ' ' + testAuditMessage
+        client.write messagePrependlength
+
```
rcrichton: Have you tried a `client.end()` here. It may make all the difference.

Also, you aren't waiting for the data to be written to the socket here so maybe that is an issue? Perhaps try end the sockets the listen for the `end` event then only check for audits. You may be looking for audits before they have been saved.

test/integration/auditingIntegrationTests.coffee
rcrichton: Also, check that are ever get into this callback without an error. If you do we can rule out incorrect certificates.

src/server.coffee
BMartinos: Cool made this into a function to reduce duplicated code

test/integration/auditingIntegrationTests.coffee
BMartinos: ive updated the code to read 'tls.connect...' or 'net.connect...'. Ive also added an .on error check which brings back connection refused. Does this mean the certificate is incorrect or it cant reach the server?

src/server.coffee
```diff
@@ -260,7 +264,58 @@ startAuditUDPServer = (auditUDPPort, bindAddress) ->
...
+        # remove white spaces
+        length = parseInt(lengthValue.trim())
+        
+        # update message to remove length - add one extra character to remove the space
+        message = message.substr(lengthIndex + 1)
+      
+      # if sourced length equals message length then full message received
+      if length == message.length
+        logger.info "[Auditing #{type}] Received message from #{sock.remoteAddress}"
+        auditing.processAudit message, -> logger.info "[Auditing #{type}] Processed audit"
```
rcrichton: Here you may want to reset `message` and `length` so that if you receive another audit on the same connection you are ready to process it.

rcrichton:
@BMartinos this looks good to me, just one comment above.

BMartinos:
@rcrichton ive made the final changes. Can you just look over this again please. Thanks

rcrichton:
Thanks, looks good to me :+1:

PR #439, opened 2015-03-13
Performance optimizations

### LINKED ISSUE #428: Use ensureIndex rather than createIndex in visualizer?
From looking at the flame graph it seems we spend some time creatIndex. It seems this is coming from https://github.com/jembi/openhim-core-js/blob/master/src/middleware/visualizer.coffee#L141

Perhaps use ensureIndex instead? http://mongodb.github.io/node-mongodb-native/markdown-docs/indexes.html#ensure-indexes-with-ensureindex

### LINKED ISSUE #429: Index user.name on clients
This may improve performance on basic authentication: https://github.com/jembi/openhim-core-js/blob/master/src/middleware/basicAuthentication.coffee#L40

### LINKED ISSUE #430: Index client.clientDomain
This may improve performance on basic authentication: https://github.com/jembi/openhim-core-js/blob/master/src/middleware/tlsAuthentication.coffee#L74

### LINKED ISSUE #431: Try not to request keystore on every request
It would be better to not fetch the keystore on every request. Is there a better way to do this while still being responsive to changes in the db?

Perhaps by only requesting once every 5 seconds?

See https://github.com/jembi/openhim-core-js/blob/master/src/middleware/router.coffee#L84

### LINKED ISSUE #432: Try not to request all channels on every request
It would be better to not fetch all channels on every request. Is there a better way to do this while still being responsive to changes in the db?

Perhaps by only requesting once every 5 seconds?

See https://github.com/jembi/openhim-core-js/blob/master/src/middleware/authorisation.coffee#L83

devcritter:
Closes #432 
Closes #431 
Closes #430 
Closes #429
Closes #428 

@rcrichton mind taking a look? Thanks

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -92,6 +92,6 @@
   },
   "caching": {
     "enabled": true,
-    "refreshMillis": 5000
+    "refreshMillis": 1000
   }
 }
--- a/config/default.json
+++ b/config/default.json
@@ -89,5 +89,9 @@
         "port": 5052
       }
     }
+  },
+  "caching": {
+    "enabled": true,
+    "refreshMillis": 5000
   }
 }
--- a/src/utils.coffee
+++ b/src/utils.coffee
@@ -19,7 +19,7 @@ refreshMillis = config.caching.refreshMillis
 getCachedValues = (store, callback) ->
   lastCheck = cacheValueStore["#{store}"]?.lastCheck
 
-  if not config.caching.enabled or not lastCheck or ((new Date)-lastCheck) > refreshMillis
+  if not config.caching.enabled or not lastCheck? or ((new Date)-lastCheck) > refreshMillis
 
     handler = (err, results) ->
       return callback err if err
--- a/src/utils.coffee
+++ b/src/utils.coffee
@@ -1,7 +1,48 @@
 logger = require 'winston'
+config = require './config/config'
+config.caching = config.get('caching')
+Channel = require("./model/channels").Channel
+Keystore = require("./model/keystore").Keystore
+
 
 # function to log errors and return response
 exports.logAndSetResponse = (ctx, status, msg, logLevel) ->
   logger[logLevel] msg
   ctx.body = msg
   ctx.status = status
+
+
+cacheValueStore = {}
+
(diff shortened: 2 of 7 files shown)


src/utils.coffee
```diff
@@ -1,7 +1,48 @@
...
+
+
+cacheValueStore = {}
+
+refreshMillis = config.caching.refreshMillis
+
+getCachedValues = (store, callback) ->
+  lastCheck = cacheValueStore["#{store}"]?.lastCheck
+
+  if not config.caching.enabled or not lastCheck or ((new Date)-lastCheck) > refreshMillis
```
rcrichton: Minor: Here should we rather do a `... not lastCheck? or ...`

config/default.json
```diff
@@ -89,5 +89,9 @@
         "port": 5052
       }
     }
+  },
+  "caching": {
+    "enabled": true,
+    "refreshMillis": 5000
```
rcrichton: Maybe we should set this to 1000 by default to ensure people generally get what they expect when using the UI?

rcrichton:
Cool looks good :+1:, just some minor comments above.

rcrichton:
Any noticeable differences in the benchmarks?

src/utils.coffee
devcritter: yeah, good idea

config/default.json
devcritter: I found that 5s goes so quickly just playing around with it on my local system. I was even tempted to up this to 10s :) But sure, yeah, let's put this on 1s by default - minimize weird results for basic users.

devcritter:
I haven't run the bench suite yet, but I did notice a small difference when running loadtest manually.

config/default.json
rcrichton: Ok cool. I think 1s will make a pretty big difference already.

rcrichton:
Well seems to have made a big difference on my machine! 270 rps on highly concurrent get. 224 on highly concurrent post.

Also, the flamechart is looking better than before in js-land.

devcritter:
Oh wow! Awesome :)

PR #441, opened 2015-03-17
Enables an option to allow clients to be found for any domain in the certification path OHM-114

### LINKED ISSUE #418: Validate clients domain against any certificate in the certification path
In large deployments it may be difficult to setup clients for hundreds of PoC systems. Therefore it would be beneficial to be able to validate a client on their issuer rather than just the client certificate. In that case the issuer cert would be uploaded to the keystore and any clients with certs signed by that authority would be valid.

rcrichton:
Closes #418 

@devcritter could you please review this for me? @bausmeier I'm not sure if you want to take a look as well.

I'm not too happy about having a dependency on the [pem](https://github.com/andris9/pem) modules master branch. I wonder if we should wait for a release to merge this in. Thoughts?


README.md
```diff
@@ -77,11 +77,12 @@ This starts the server with production defaults, including the use of the produc
 
 This project uses [mocha](http://visionmedia.github.io/mocha/) as a unit testing framework with [should.js](https://github.com/visionmedia/should.js/) for assertions and [sinon.js](http://sinonjs.org/) for spies and mocks. The tests can be run using `npm test`.
 
-**Useful tips:**
+**Pro tips:**
 
 * `grunt watch` - will automatically build the project on any changes.
 * `grunt lint` - ensure the code is lint free, this is also run before an `npm test`
 * `npm link` - will symlink you local working directory to the globally installed openhim-core module. Use this so you can use the global openhim-core binary to run your current work in progress. Also, if you build any local changes the server will automatically restart.
+* `grunt test --mochaGrep=<regex>` - will only run tests with names matching the regex
```
devcritter: nice!

devcritter:
:+1: Yeah, looks good to me. It took me a bit to fully understand; but it makes sense now. The `clientLookup` method is really elegant, nice one!

On the pem module master branch: I'm actually okay with having that in master. I would rather just say that we should wait before doing a release.

rcrichton:
Thanks @devcritter. It was rather confusing to write to! :) - I've added some better comments.

I've added a few fixes and better testing from some of the issue @bausmeier chatted to me about.

Could you guys have a quick look again?

devcritter:
:+1: Looks good to me

bausmeier:
:+1:

PR #442, opened 2015-03-17
Tasks batching strategy

### LINKED ISSUE #434: Allow tasks to be paused/resume

### LINKED ISSUE #435: Allow task worker thread to be enabled/disabled in config
The would useful ito scalability so that certain nodes can be dedicated to rerunning tasks

### LINKED ISSUE #438: Tasks: multiple transaction running strategies
Currently when a task runs, the worker thread will move on to the next item in the task as soon as the request for the previous item has been fired off (i.e. it doesn't wait for a transaction to finish first). This is definitely a good strategy as it keeps the worker busy, but it can quickly overwhelm a target server with too many transactions.

For certain contexts it would be better to run transactions one by one, waiting for each to finish before moving on to the next.

Taking into account that sometimes you do want to move through a task faster than one at a time, should we introduce a configurable running strategy? Perhaps when submitting a task, along with the transaction ids, maybe you could specify a concurrency level for that task?

devcritter:
Closes #438 
Closes #435 
Closes #434 

@rcrichton could you please review?

Note I've replace monq with a custom processor. The model works a bit differently now, only the tasks collection is used to manage reruns and there is no separate queue.

When a task is first inserted, it's marked as queued. When "dequeued" (i.e. picked up and marked as "Processing"), tasks are processed in rounds where a `batchSize` number of transactions are ran. After completing the round, the task is then again marked as queued and made available to be picked up again.

This is more scalable as multiple core instances can then process tasks (different tasks are processed in parallel, but there'll never be more than a batchSize number of transactions running at a moment for a specific task). The tasks API now only inserts tasks in the collection, marked as Queued, but doesn't do anything else. The rerun process (or processes on multiple instances) simply monitor the collection for queued items.

It's also more robust as it allows for graceful shutdowns - between rounds we can check whether the task has been paused, cancelled or if the server has been stopped; but since the check is done between rounds, currently running transactions will first finish before stopping a task.

@BMartinos @debonair since you guys done a lot of work on this, it would be great if you could take a look as well to see if it makes sense?


rcrichton:
Hey @devcritter this sounds great! Just one question (I'm just started to go through the code, so I may find the answer in the code). As mongo is eventually consistent is it not a possibility that two process pickup the same set of transaction as the update hasn't hit the node they are reading from yet?

devcritter:
Hmmm, that's a good question - I haven't really considered that. Maybe @armageddon has some ideas?

src/tasks.coffee
```diff
@@ -0,0 +1,314 @@
...
+  setTimeout rerunTaskProcessor, config.rerun.processor.pollPeriodMillis
+
+  logger.info "Started rerun task processor"
+  callback()
+
+
+exports.stop = (callback) ->
+  live = false
+
+  waitForActiveTasks = ->
```
rcrichton: It may be useful to use a array of promises for running tasks rather than setting a timeout. Not sure how feasible this is.

src/tasks.coffee
```diff
@@ -0,0 +1,314 @@
...
+
+      rerunTransaction transaction.tid, task._id, (err, response) ->
+        if err
+          transaction.tstatus = 'Failed'
+          transaction.error = err
+          logger.error err
+        else if response?.status is 'Failed'
+          transaction.tstatus = 'Failed'
+          transaction.error = response.message
+          logger.error response.message
```
rcrichton: Maybe log a bit more context for the previous 2 logger.error messages. Ie. what you are logging out.

src/tasks.coffee
```diff
@@ -0,0 +1,314 @@
...
+
+
+#####################################
+# Construct HTTP options to be sent #
+#####################################
+
+rerunSetHTTPRequestOptions = (transaction, taskID, callback) ->
+
+  if transaction == null
+    err = "An empty Transaction object was supplied. Aborting HTTP options configuration"
```
rcrichton: I think its best to construct a new Error object for this sort of thing.

src/tasks.coffee
```diff
@@ -0,0 +1,314 @@
...
+  promises = []
+  nextI = task.transactions.length - task.remainingTransactions
+
+  for transaction in task.transactions[nextI ... nextI+task.batchSize]
+    do (transaction) ->
+      defer = Q.defer()
+
+      rerunTransaction transaction.tid, task._id, (err, response) ->
+        if err
+          transaction.tstatus = 'Failed'
```
rcrichton: I'm not sure I understand this transaction objects use. It doesn't seem to be saved and a task only stored tids doesn't it?

BMartinos:
Looks awesome to me! Works really well :)

rcrichton:
@devcritter perhaps we have to do something with read/write concerns?

src/tasks.coffee
```diff
@@ -0,0 +1,314 @@
...
+          return callback err if err
+
+          # Update original
+          ctx =
+            parentID : transaction._id
+            transactionId : transactionID
+            transactionStatus: TCPResponse.status
+            taskID : taskID
+
+          rerunMiddleware.updateOriginalTransaction ctx, (err) ->
```
rcrichton: I don't quite get why the middleware needs to be called here, surely its called when rerunning each transaction ie. https://github.com/jembi/openhim-core-js/blob/master/src/koaMiddleware.coffee#L99-L100

I understand this may be copied code, but still it seems redundant.

rcrichton:
Hey @devcritter looks very good. Just a few comment above. Perhaps its me struggling to get my head around this.

src/tasks.coffee
devcritter: Not sure I follow - So the setTimeout triggers the polling process. We could use promises to check for active tasks, rather the counter; is that what you mean? It might just be a bit tricky to do it that way right now, and this works.

devcritter:
I just chatted to Pierre about this and he pointed out that since writes are only allowed on the primary, it shouldn't be a problem because we're doing "dequeueing" using a [findAndModify](https://github.com/jembi/openhim-core-js/blob/tasks-batching-strategy/src/tasks.coffee#L18), so different processes should always be working from the primary.

Additionally he suggested we can also set a write concern for these calls to only work off the primary - we can just that it is working properly with the findAndModify

src/tasks.coffee
rcrichton: Yeah, that's what I mean. So instead of activeTasks we have a managed array of promises. Don't worry about this too much, just through it may be a better way but it may add complexity to. I'm ok with either way.

src/tasks.coffee
devcritter: No, we're storing more than just the tid (https://github.com/jembi/openhim-core-js/blob/master/src/model/tasks.coffee#L8), we're storing things like the transaction process status and result. The transaction object is part of the task object, so these updates take effect [here](https://github.com/jembi/openhim-core-js/blob/tasks-batching-strategy/src/tasks.coffee#L80) and [here](https://github.com/jembi/openhim-core-js/blob/tasks-batching-strategy/src/tasks.coffee#L120)

rcrichton:
Ok cool, that sound good then. So, when working with the primary is the doc not about to be read while updating?

src/tasks.coffee
devcritter: The TCP rerunning code doesn't use the rerunApp middleware - it goes straight to the original socket. But I think see some other issues with this, namely that it's not using the correct [transaction id](https://github.com/jembi/openhim-core-js/blob/tasks-batching-strategy/src/tasks.coffee#L150-L151). Our design for TCP might need some rethinking - maybe it should be using that middleware.

But I think this is out of scope for this PR - we should probably pick this up on another ticket?

src/tasks.coffee
rcrichton: Ah ok, I see. I was getting confused with this and the actual transaction model.

src/tasks.coffee
rcrichton: Ok, I see. Yeah this can happen in another issue. It just seems rather weird now that I'm seeing it again.

devcritter:
[Looks](http://docs.mongodb.org/manual/reference/command/findAndModify/) like it's a single operation - so I think we should be fine: the first process to find a queued task would immediately mark it as "taken" - i.e. status=Processing.

src/tasks.coffee
devcritter: Cool sure, yeah, I'm just a little worried about the complexity - just with the way it's setup atm. But yeah we should definitely revisit if the current way causes issues. It'll be easy to tell when that is -  core won't shutdown even though there aren't any running transactions.

src/tasks.coffee
devcritter: Yeah it's because updating per reference - always tricky to read :/

src/tasks.coffee
devcritter: Good point - will add.

devcritter:
@rcrichton thanks for the feedback. I've made relevant changes.

rcrichton:
Cool, seems we are safe using a [findAndModify](http://docs.mongodb.org/manual/faq/concurrency/#what-type-of-locking-does-mongodb-use)

Looks good to me :+1:

devcritter:
Aah okay, great

PR #445, opened 2015-03-19
OHM-117: Allow a task to be added in a 'Paused' state

devcritter:
@rcrichton a quick additional feature following #434 - this allows a task to be added in 'paused' state so that it doesn't begin running immediately.


rcrichton:
:+1:

PR #446, opened 2015-03-19
Transaction advanced filtering

BMartinos:
This PR relates to issue: https://github.com/jembi/openhim-console/issues/256


BMartinos:
Hey @rcrichton Do you mind just reviewing this PR for me when you get a chance. Thanks

src/api/transactions.coffee
```diff
@@ -40,19 +41,21 @@ getProjectionObject = (filterRepresentation) ->
 ###
 # Retrieves the list of transactions
 ###
+
+# function to check if string match status code pattern
+patternMatch = (string) ->
+  pat = /\dxx/.exec string
```
rcrichton: I think you may want to use `.test` here

src/api/transactions.coffee
```diff
@@ -40,19 +41,21 @@ getProjectionObject = (filterRepresentation) ->
...
+  if patternMatch
```
rcrichton: I think you mean to use `pat` here.

src/api/transactions.coffee
```diff
@@ -317,3 +393,28 @@ exports.removeTransaction = (transactionId) ->
...
+
+
+
+###
+# construct transaction filtering dropdown options
+###
+exports.getTransactionsFilterOptions = ->
+
+  # Must be admin
+  if not authorisation.inGroup 'admin', this.authenticated
```
rcrichton: non-admins may need this function to. You may need to search for only the channels and clients that a user is authorised to view.

src/api/transactions.coffee
```diff
@@ -317,3 +393,28 @@ exports.removeTransaction = (transactionId) ->
...
+    return
+
+  try
+    # execute the query
+    channels = yield Channel.find().exec()
+    clients = yield Client.find().exec()
+
+    responseObject =
+      channels: channels
+      clients: clients
```
rcrichton: If you only need channels and clients could you not rather use the two API call to fetch those using the existing API calls?

src/koaApi.coffee
```diff
@@ -58,6 +58,7 @@ exports.setupApp = (done) ->
   app.use route.get '/transactions/clients/:clientId', transactions.findTransactionByClientId
   app.use route.put '/transactions/:transactionId', transactions.updateTransaction
   app.use route.delete '/transactions/:transactionId', transactions.removeTransaction
+  app.use route.get '/transactions-filter-options', transactions.getTransactionsFilterOptions
```
rcrichton: Perhaps  rename this to `transactionFilterOptions` to be more consistent.

test/integration/transactionsAPITests.coffee
```diff
@@ -272,14 +273,32 @@ describe "API Integration Tests", ->
                   done()
 
       it "should call getTransactions with filter paramaters ", (done) ->
```
rcrichton: It would be good if there was a bit more comprehensive test coverage of the different filter param options.

rcrichton:
Hey @BMartinos, some comments above.

src/api/transactions.coffee
BMartinos: Yes you are right. this slipped past me. Will add a test for this

src/api/transactions.coffee
BMartinos: Will update this check to return appropriate channels

src/api/transactions.coffee
BMartinos: I originally had it as two API calls but though it might be better to make one call and return all the filters needed. Reason for this is that it would be easier to add more filter options to this return if needed in the future

rcrichton: I'd rather not add extra API call if we don't have to. I want to keep the API as lean and mean as possible.

BMartinos:
Hey @rcrichton I have completed the changes you recommended. I have also removed the additional API endpoint. The console will now make two calls to get the clients and channels

rcrichton:
Thanks, this look good @BMartinos :+1:

PR #451, opened 2015-03-20
Cluster API integration to allow TCP ports to be shared

### LINKED ISSUE #405: The design of the TCP channel adapter stops us for being able to scale horizontally

rcrichton:
Closes #405 

This PR allows the OpenHIM to use the cluster API. This will share all servers with the workers, both TCP and HTTP, so,  this allows us to cluster even with a single config file.

There is a bug in node preventing me from clustering the UDP sockets (See joyent/node#9261) so those don't get clustered yet. The first worker to get to the server up hosts it.

Use as follows: `node --harmony lib/server/js --cluster=auto` - or replace auto with a number of workers.

@devcritter could you please review this?

### DIFF
--- a/src/server.coffee
+++ b/src/server.coffee
@@ -34,8 +34,7 @@ tlsAuthentication = require "./middleware/tlsAuthentication"
 uuid = require 'node-uuid'
 Q = require 'q'
 logger = require 'winston'
-logger.level = config.logger.level
-logger.colorize = true
+logger.remove logger.transports.Console
 cluster = require 'cluster'
 numCPUs = require('os').cpus().length
 nconf = require 'nconf'
@@ -54,7 +53,16 @@ tasks = require './tasks'
 clusterArg = nconf.get 'cluster'
 
 # Configure clustering if relevent
-if cluster.isMaster and clusterArg?
+if cluster.isMaster
+
+  # configure master logger
+  logger.add logger.transports.Console,
+    colorize: true
+    label: "master"
+    level: config.logger.level
+
+  if not clusterArg?
+    clusterArg = 1
 
   if clusterArg is 'auto'
     clusterSize = numCPUs
@@ -93,6 +101,12 @@ if cluster.isMaster and clusterArg?
   cluster.on 'listening', (worker, address) ->
     logger.debug "worker #{worker.process.pid} is now connected to #{address.address}:#{address.port}"
 else
+  # configure worker logger
+  logger.add logger.transports.Console,
+    colorize: true
+    label: "worker#{cluster.worker.id}"
+    level: config.logger.level
+
   # Configure mongose to connect to mongo
   mongoose.connect config.mongo.url
 
@@ -515,7 +529,6 @@ else
       # restart on message
       process.on 'message', (msg) ->
         if msg is 'restart'
-          console.log 'calling restart'
           restartServer()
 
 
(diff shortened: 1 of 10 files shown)


devcritter:
Not sure if it's related to the udp bug - but I get the following error occasionally when stopping the server:
```
Error: Not running
    at Socket._healthCheck (dgram.js:452:11)
    at Socket.close (dgram.js:353:8)
    at /openhim-core-js/src/server.coffee:486:22
    at stopTasksProcessor (/openhim-core-js/src/server.coffee:461:7)
    at process.exports.stop.stop (/openhim-core-js/src/server.coffee:463:35)
    at process.emit (events.js:107:17)
    at process.exit (node.js:600:17)
    at process.<anonymous> (cluster.js:532:17)
    at process.g (events.js:199:16)
    at process.emit (events.js:104:17)
```

Maybe you should catch that error? (I couldn't see a `isClosed` method in the api)

src/server.coffee
```diff
@@ -14,6 +17,7 @@ config.alerts = config.get('alerts')
 config.polling = config.get('polling')
 config.reports = config.get('reports')
 config.auditing = config.get('auditing')
+config.agenda = config.get('agenda')
 
 mongoose = require "mongoose"
 exports.connectionDefault = connectionDefault = mongoose.createConnection(config.mongo.url)
```
devcritter: Do the mongo connections still need run if the node is master? I believe this will start up a connection pool, so it might save some resources by excluding this

src/server.coffee
```diff
@@ -44,438 +51,526 @@ tcpAdapter = require './tcpAdapter'
...
+    logger.debug "worker #{worker.process.pid} is now connected to #{address.address}:#{address.port}"
+else
+  # configure worker logger
+  logger.add logger.transports.Console,
+    colorize: true
+    label: "worker#{cluster.worker.id}" if cluster.worker?.id?
+    level: config.logger.level
+
+  # Configure mongose to connect to mongo
+  mongoose.connect config.mongo.url
```
devcritter: Out of scope question :) Is this still needed with the createConnection calls above?

devcritter:
Found an issue when creating new tcp channels from the console - it only starts up the tcp server on one of the workers. When restarting core though, all the workers pick up the tcp channel correctly. Presumably this happens because it's starting a tcp socket on whichever worker gets the API call.

A quick solution might be for the API call to just trigger a general restart?

devcritter:
(the same issue with stopping a TCP channel, e.g. deleting it)

devcritter:
@rcrichton Just some comments above, but otherwise it works really well! Very cool :)

src/server.coffee
rcrichton: Good point. I'l look at removing it.

src/server.coffee
rcrichton: Honestly I don't know. It was already in the code when I made these changes. I'l look into it though.

rcrichton:
Hey, @devcritter I also get that dgram server error some times. I don't think is an issue due to clustering. However, it may be because it isn't possible to wait for the dgram server to close as the function doesn't have a callback (the node API seems a bit different to the norm of the tcp and http API here). I will try to catch this but I'm not quite sure where it's getting thrown, I'm guess at server.close() time.

Also, good call on the TCP channels. I'l have to do some more work there.

src/server.coffee
rcrichton: It seems it isn't needed.

src/server.coffee
rcrichton: Hey @devcritter, it only possible to remove these if we move all `/model/*` requires out of the master as these exported connection [are](https://github.com/jembi/openhim-core-js/blob/master/src/model/channels.coffee#L79) [used](https://github.com/jembi/openhim-core-js/blob/master/src/model/transactions.coffee#L55-L57) [there](https://github.com/jembi/openhim-core-js/blob/master/src/model/audits.coffee#L60). I'm not too keen to make that change as its rather large, I don't know if it will affect anything else and I don't think we'll save much by doing so. What do you think?

rcrichton:
Hey @devcritter, I've addressed the comments. Could you have a look at this again?

src/server.coffee
devcritter: Aah okay, yeah if it's a mission than definitely skip it. I'm sure it's not too big a deal ito resource usage.

devcritter:
:+1: Looks good to me. Excited to see this merged in! :)

Just note that the failing tests before merging

rcrichton:
Cool. The test are failing due to a bad mongoose version that I fixed in master this morning. I've merged in master here so they should pass now. I'l wait until they pass before merging this.

PR #454, opened 2015-03-24
Tasks pagination

BMartinos:
This PR relates to issue: https://github.com/jembi/openhim-console/issues/255

Added filtering and pagination additions to the getTask() and getTasks() API function. Updated tests. Found a bug in src/tasks.coffee that broke the server

### DIFF
--- a/src/api/tasks.coffee
+++ b/src/api/tasks.coffee
@@ -57,7 +57,7 @@ exports.getTasks = ->
 
     #determine skip amount
     filterSkip = filterPage*filterLimit
-    
+
     # get filters object
     filters = JSON.parse filtersObject.filters
 
@@ -67,6 +67,10 @@ exports.getTasks = ->
 
     # exclude transactions object from tasks list
     projectionFiltersObject = { 'transactions': 0 }
+
+    console.log( filters )
+
+    this.body = yield Task.find({}).exec()
 
     # execute the query
     this.body = yield Task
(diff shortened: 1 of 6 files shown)


BMartinos:
Hey @devcritter Do you mind reviewing this PR for me when you get a chance. Thanks

src/api/tasks.coffee
```diff
@@ -128,12 +158,71 @@ exports.getTask = (taskId) ->
...
+    filterSkip = filterPage*filterLimit
+    
+    # get filters object
+    filters = JSON.parse filtersObject.filters
+
+    result = yield Task.findById(taskId).lean().exec()
+    tempTransactions = result.transactions
+
+    # are filters present
+    if Object.keys( filters ).length > 0
```
devcritter: This method's getting a little big and it might be a good idea to split it up into smaller units of work. This section looks like a good candidate. This could easily be split into a method `buildFilteredTransactionsArray` or something similar.

test/integration/tasksAPITests.coffee
```diff
@@ -326,8 +342,23 @@ describe "API Integration Tests", ->
     describe '*getTask(taskId)', ->
```
devcritter: It might be good to add some more tests for getTask: two examples I can think of are:
* Add a task with failed and completed transactions, and see if the filter only returns the failed transactions
* Add a task with >10 transactions and check that the returned number is 10 (with a limit turned specified)

devcritter:
Thanks @BMartinos! It looks great. Just some comments above.

BMartinos:
Hey @devcritter I have added some more test coverage for the filtering and moved the transaction array filtering to a function. Mind just reviewing this for me again. Thanks

devcritter:
:+1: Thanks that looks great

PR #455, opened 2015-03-25
Check that updates.$push is defined

bausmeier:
Prevent the HIM from crashing when updating a transaction with some operation other than a push to the routes array.


rcrichton:
Looks good to me :+1:

PR #462, opened 2015-03-27
OHM-61: Replaced http status strings with codes for koa compatibility

### LINKED ISSUE #318: Update koa and fix breaking changes

devcritter:
Closes #318 

@rcrichton if you could please take a look? @bausmeier it would be useful if you could take a look as well, seeing as you picked this up originally?

Replaced all the status string with codes. Used ack to figure out what statuses we're using (`ack --type coffeescript -o 'status ?= ?(.*)'`) and a combination of ack and perl to substitute the matches (using the approach outline in the first answer [here](http://stackoverflow.com/questions/4792561/how-to-do-search-replace-with-ack-in-vim)).

I *think* I've got everything. The tests are passing and my test transactions are happy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "forever-monitor": "^1.5.2",
     "glossy": "^0.1.7",
     "http-status": "^0.1.8",
-    "koa": "^0.7.0",
+    "koa": "0.18.1",
     "koa-basic-auth": "^1.1.2",
     "koa-body-parser": "^1.1.0",
     "koa-compress": "^1.0.8",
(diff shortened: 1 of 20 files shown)


bausmeier:
Looks good to me. Can't find any that you missed.

rcrichton:
Looks good thanks. My search returned 0 that you missed. :+1:

Issue #354, opened 2015-02-06
Enable the OpenHIM to act as a CA


rcrichton:
The OpenHIM should be able to act as a CA. This mean that we should be able to generate certificates for clients on the fly. Also, the OpenHIM should allow the admin to generate a self-signed certificate for the OpenHIM server instance itself. This certificate should have the ability to sign other certificates. This mean that is must have the basicConstrainsts extension cA flag set to true (http://tools.ietf.org/html/rfc5280#section-4.2.1.9). This can cause problem as the there is no known module to do this in node land. Two possible options are:

* Use a openssl wrapper: https://www.npmjs.com/package/openssl-wrapper
* Modify pem to support this: https://www.npmjs.com/package/pem

Also, we would need a way to check that a certificate that a user uploads as the HIM server cert support signing certificates. If it does not we should disable the ability to generate client certificates.

debonair:
@rcrichton What should the form on the console side look like? I have pushed an initial version for you to take a look. I want to know what you expect to happen after user fills in the form, ie what the API should return as a result.

Other questions:

1. Should I save the result to mongo
2. what files should I return the certificate for the user to download

rcrichton:
Hey @debonair, the UI should allow the user to generate a certificate for a client (which is then added to the trusted certs list). They should be able to generate a certificate for the OpenHIM server itself. I'd guess you need buttons to initiate these actions then a form to capture the certificate details (common name, organisation, organisation unit etc. and the validity period). I think the API should just return 201 if the creation of the cert was successful. For generated client certs we should allow the user to download the cert and key, but I think this should only be allowed directly after generating the cert to prevent us from having to store the key for security reasons.

@devcritter @bausmeier thoughts?

devcritter:
Should we rather use CSR's to sign client certificates, rather than generating the private key ourselves? Not that I'm a PKI expert at all, but it appears to be the standard way for CAs to operate. Of course we could still have functionality to generate the certificates/keys, it would certainly be useful, but currently we would exclude any clients that want to generate their own certs/keys and have those signed by the HIM.

rcrichton:
I think that our first goal should be to make this as easy as possible for people setting up an OpenHIM instance. Generating CSRs isn't that straight forward if you are new to it. So I think we should do everything via the UI. That said, I agree, CSRs are the 'better' way to do this. Perhaps we can add support for this as well. I think we could create a new feature request for this so that we don't add too much complexity into this issue.

[...]

PR #464, opened 2015-03-27
Him as ca

debonair:
Added functionality to generate and add a certificate to the trust store


src/api/certificateAuthority.coffee
```diff
@@ -0,0 +1,56 @@
...
+utils = require "../utils"
+pem = require "pem"
+
+exports.createCert = ->
+  certParams = this.request.body
+  keystoreDoc = yield Keystore.findOne().exec()
+  options = certParams
+
+  #This flag determines if this is a CA cert or a client cert
+  if !certParams.selfSigned
```
rcrichton: Perhaps we should split this out into two different functions. I think we 'll need something like generateServerCert and generateClientCert api calls. This is also because we would want to do different things with the client certs as opposed to the server cert. The client cert we would save to `keystore.ca` whereas the server cert we would save to `keystore.key` and `keystore.cert.data`. We would also need two different button on the console for these two different functions to.

rcrichton:
Hey @debonair so far I think that you are on the right path. I've added a comment above with some more thoughts.

debonair:
@bausmeier @devcritter please review this pull request

devcritter:
@debonair sure will take a look. In the meantime I'd also recommend @rcrichton takes a look if he's got time? Afraid I'm a bit out of the loop on this.

devcritter:
:+1: Thanks @debonair looks good to me!

rcrichton:
Had a retrospective look, this looks good to me. Thanks @debonair.

Just one thing we should think about. I see we aren't using the HIM certificate to sign the client certificates. I think this is ok as we add the certificate directly to the keystore.ca array and don't need to  use the server cert to check validity. I don't foresee any problem with this, I just wanted to raise it so others can think about it and see if we are missing something by not doing this.

bausmeier:
@rcrichton @debonair There is security issue with the way we decided to do this. The problem is that the self-signed certs that the HIM is generating can be used to sign other certs. This means that if a cert is issued for one client's domain they can then use the cert to generate another cert for a different client's domain. This cert can then be used to access the other client's channels and will then be trusted by the HIM because it is signed by a trusted cert.

Duma and I discussed this yesterday and didn't think it would be possible, but I have tested it and it is definitely a problem.

So I think we need to either see if there is a way to issue a self-signed cert that can't be used to issue other certs, or we need to go back to issuing certs signed by the HIM cert. The original problem with the second way is that we then don't have any method for revoking certs.

What do you guys think?

rcrichton:
@bausmeier thanks that is definitely a problem. I think the best way to get around this would be to try make pem not generate certificates that have can be used to sign other certs. Pem doesn't seem to set any options to do this that I can see, perhaps this is because we have `basicConstraints=CA:TRUE` in our default openssl config files?

rcrichton:
Brett, made a good point. Self signed certs need to be able to sign as they need to sign themselves. So my suggestion above isn't valid.

I'm not really sure how much of an issue this is though as you would need the key of the self signed client cert to sign another cert and this should be well protected. Its basically the same thing as having someone steal the self signed cert key. So, I'm not sure there is really an issue here that we can protect against.

rcrichton:
So, I was wrong again in chatting with Brett. The problem is a client can use their key and cert to sign a new certificate with a different commonName which would give them access to other client channels. They can basically modify their privileges.

However, this isn't an issue introduced by this PR it actually something that will affect any self-signed cert that we trust. I'l create another issue to capture this and link it here for us to deal with.

debonair:
closes #354

PR #471, opened 2015-04-01
Uptime endpoint

BMartinos:


BMartinos:
Hey @rcrichton Do you mind just reviewing this PR for me. I have some merge conflicts but will sort them out now.

BMartinos:
Credit to @bausmeier for teaching me more about event listerners

src/server.coffee
```diff
@@ -80,6 +80,12 @@ if cluster.isMaster and not module.parent
     worker = cluster.fork()
 
     worker.on 'message', (msg) ->
+
+      if msg.uptime
```
rcrichton: Could you perhaps make msg have a .type property just so that is conforms to the way the other messages are done below?

src/server.coffee
```diff
@@ -591,3 +597,24 @@ else
...
+      uptime =
+        master: process.uptime()
+      callback null, uptime
+    else
+      # send request to master
+      process.send
+        uptime: true
+
+      # listen for response from master
+      process.once 'message', (uptime) ->
```
rcrichton: There is a minute chance that a server restart would be called while this code is waiting for a uptime response, this would cause it to get a restart message instead of a uptime response message. Its really unlikely so I don't think we have to change for using .once here but maybe we should add a .type property here to check that it is of the correct type before trying to call the callback?

rcrichton:
Hey @BMartinos looks great! Just a few small comments.

BMartinos:
Hey @rcrichton i have made those changes as mentioned. Can you review this again for me please. Thanks

src/koaApi.coffee
```diff
@@ -19,6 +19,8 @@ serverRestart = require './api/restart'
 audits = require './api/audits'
 statsd = require './api/statsd'
 config = require './config/config'
+heartbeat = require './api/heartbeat'
+#statsd_instance = config.get 'statsd'
```
rcrichton: I think this commented code can be deleted, not sure where it came from.

src/server.coffee
```diff
@@ -591,3 +597,25 @@ else
...
+        master: process.uptime()
+      callback null, uptime
+    else
+      # send request to master
+      process.send
+        type: 'get-uptime'
+
+      # listen for response from master
+      process.once 'message', (uptime) ->
+        if uptime.type is 'get-uptime'
```
rcrichton: I just thought of something, perhaps we should add an else statement here and call the callback with:

```
{
  master: null
}
```

This would be so that the API call would respond even if it got a message it wasn't expecting.

rcrichton: Maybe a quick null check in api/heartbeat.coffee and returning a 501 would be useful to?

rcrichton:
Hey @BMartinos looks great. Just a few more comments that I just thought of.

src/server.coffee
BMartinos: If the event listener goes into the 'else' then the request wouldnt have come from the heartbeat API request. Would this then ever return null to the heart API call?

rcrichton: The main issue I'm thinking of is that the `.once` off message listener will be cancelled if it gets a message that it isn't expecting (ie. a restart message, or something else in the future). So, what about we do something like this:

```
# listen for response from master
setupListener = ->
  process.once 'message', (uptime) ->
    if uptime.type is 'get-uptime'
      uptime =
        master: uptime.masterUptime

      # send reponse back to API request
      callback null, uptime
    else
      setupListener()

setupListener()
```

This will resetup the listener if we don't get the message we were expecting.

bausmeier: Would it not be better to use `on` and remove the listener when the uptime response is received?

BMartinos: That could also work. so we only process the callback when the type is 'get-uptime' (ignore anything else) and then we destroy the listener after that

rcrichton: @bausmeier yeah that's fine to.

Issue #466, opened 2015-03-30
Daily/Weekly reports not being sent


rcrichton:

debonair:
There is a  need to update agenda to the latest version currently on version on 0.6.26
the latest is 0.6.28. There have been some bug fixes since then.

rcrichton:
Is this causing the issue though? There are probably only minor bug fixes in the new release.

debonair:
The issue we found was that the jobs were locked. @bausmeier has released the locks, we will see if we get a daily report tomorrow. I still guess it is prudent to update to the latest agenda if we can. I could not figure out why the jobs were locked in the first place. I am looking into that

rcrichton:
@debonair I'm still not getting the reports. Any ideas?

debonair:
Check if the job is locked in the the agenda collection

rcrichton:
Have you been able to get any reports out of the system?

PR #474, opened 2015-04-07
fixed plaintext function

debonair:
Fixed issue with daily and weekly reports

### DIFF
--- a/src/reports.coffee
+++ b/src/reports.coffee
@@ -128,9 +128,10 @@ fetchWeeklySubscribers = (callback) ->
   User.find { weeklyReport: true }, callback
 
 plainTemplate = (report) ->
+  text = ''
   for data in report.data
     do (data) ->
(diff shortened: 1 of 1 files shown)


debonair:
@rcrichton please review the plaintext method was returning an object and not plain text so it was breaking the email send.

src/reports.coffee
```diff
@@ -197,10 +200,8 @@ setupAgenda = (agenda) ->
   agenda.every "1 days", 'send daily channel metrics'
 
 exports.setupAgenda = setupAgenda
-
-if process.env.NODE_ENV == "test"
```
rcrichton: Could you add this back in. I don't think relates to the chance that was made.

rcrichton:
Hey @debonair looks good to me. Just add back the check for the test env as mentioned above, then :+1:

src/reports.coffee
debonair: Ok cool

debonair:
closes #466

Issue #460, opened 2015-03-26
Unhandled ssl error


bausmeier:
There seems to be a missing error handler on one of the tls sockets which causes the HIM to crash occasionally. This error should be handled so that only the connection fails, not the whole HIM.

```
events.js:85
      throw er; // Unhandled 'error' event
            ^
Error: 140542554728256:error:140940E5:SSL routines:SSL3_READ_BYTES:ssl handshake failure:../deps/openssl/openssl/ssl/s3_pkt.c:1026:

    at Error (native)
```

BMartinos:
Some extra logging below:
---------------------------------------------------
praekelt is authenticated.
info: The request, '/ws/rest/v1/subscription' is authorised to access Subscription
info: Storing request metadata for inbound transaction
info: Routing http(s) request
info: Routing http(s) request
info: Storing response for transaction: 550fe63011c777002a9dcd9f
info: sending counts to statsd for momconnect-him.Production.550c2c5febaab96a71b3d4e0
info: sending durations to statsd for momconnect-him.Production.550c2c5febaab96a71b3d4e0
info: Storing response for transaction: 550fe63011c777002a9dcda0
info: sending counts to statsd for momconnect-him.Production.550c2c5febaab96a71b3d4dd
info: sending durations to statsd for momconnect-him.Production.550c2c5febaab96a71b3d4dd
info: Storing visualizer events for transaction: 550fe63011c777002a9dcd9f
info: Storing visualizer events for transaction: 550fe63011c777002a9dcda0
info: praekelt is authenticated.
info: The request, '/ws/rest/v1/facilityCheck' is authorised to access Facility Check
info: Storing request metadata for inbound transaction
info: Routing http(s) request
info: Routing http(s) request
info: Storing response for transaction: 550fe63d11c777002a9dcdb5
info: sending counts to statsd for momconnect-him.Production.550c2c5febaab96a71b3d4db
info: sending durations to statsd for momconnect-him.Production.550c2c5febaab96a71b3d4db
info: Storing visualizer events for transaction: 550fe63d11c777002a9dcdb5
info: Stopped rerun task processor
events.js:85
      throw er; // Unhandled 'error' event
            ^
Error: 140089564587840:error:140940E5:SSL routines:SSL3_READ_BYTES:ssl handshake failure:../deps/openssl/openssl/ssl/s3_pkt.c:1026:

    at Error (native)
Forever detected script exited with code 1
Forever restarting script for 1 time
[...]

PR #476, opened 2015-04-10
OHM-134: Added error checking for server and client

BMartinos:
This PR relates to issue: #460

### DIFF
--- a/src/middleware/router.coffee
+++ b/src/middleware/router.coffee
@@ -247,6 +247,8 @@ sendHttpRequest = (ctx, route, options) ->
           defered.resolve response
 
   routeReq.on "error", (err) -> defered.reject err
+  
+  routeReq.on "clientError", (err) -> defered.reject err
 
   routeReq.setTimeout config.router.timeout, -> defered.reject "Request Timed Out"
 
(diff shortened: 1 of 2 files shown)


BMartinos:
Hey @devcritter @bausmeier @rcrichton ive added in some exception listens for the server but im not sure if ive maybe missed some types or places i need to still put it.

We also noticed a weird 'stop' process happening when the error occurs ( but not always ). This is explain a bit more in the issue #460

rcrichton:
@BMartinos this looks good to me except can you use `logger.error` rather. Also, I wonder if this error is actually happening on the router (ie. outgoing connection) it may be looking for uncaught errors there to.

devcritter:
Thanks @BMartinos looks good to me as well :+1: Agree that you should just switch to log.error.

BMartinos:
Hey @rcrichton @devcritter ive updated the logger to use 'error'. also added an error listen in the router. If you can just review this for me again please. Thanks

devcritter:
Sure, looks good to me :+1:

rcrichton:
Thanks, looks good to me to :+1:

PR #480, opened 2015-04-21
Fix memory leak

bausmeier:
By comparing heap snapshots I found that references to closed `TLSSocket`s are being retained by the `activeApiConnections` object. After some investigation I found that the close listener that is being added in `trackConnection` is never called. It turns out the reference that is passed to the `connection` event on HTTPS servers is to the underlying `net.Socket` and not to the `TLSSocket` itself. For some reason the `close` event is not emitted on the underlying socket so the listener is never being called and the references aren't being cleaned up. By listening to the `secureConnection` event instead of the `connection` event we get a reference to the `TLSSocket` which does emit the `close` event.

### DIFF
--- a/src/server.coffee
+++ b/src/server.coffee
@@ -271,7 +271,7 @@ else
       httpsServer.on 'clientError', (err) ->
         logger.error "An httpsServer clientError occured: #{err}"
 
-      httpsServer.on 'connection', (socket) -> trackConnection activeHttpsConnections, socket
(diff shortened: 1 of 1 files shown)


bausmeier:
@rcrichton Please review.

rcrichton:
Thanks @bausmeier this looks great! :+1: Nice find.

psbrandt:
Very nice dude!

PR #481, opened 2015-04-22
Parse response status to an int if it is a string

bausmeier:
This is for backwards compatibility with mediators which return status codes as strings. This used to work before we upgraded koa.

@rcrichton Please take a look at this.


rcrichton:
:+1:

PR #484, opened 2015-04-24
closes #483

### LINKED ISSUE #483: Fix daily and weekly reports
Daily and weekly reports don't seen to be working. I'm done a bit of digging into this, but haven't got to the bottom of it. Here is what I found:

* It looks like a channel is associated with only one user, wouldn't this prevent reports being produced for multiple users? https://github.com/jembi/openhim-core-js/blob/master/src/reports.coffee#L38
* In my local testing it gets stuck here: https://github.com/jembi/openhim-core-js/blob/master/src/reports.coffee#L58 it seems that not all promises are being resolved. I see that no error condition are being handled, perhaps `fetchChannelReport()` isn't calling the callback on an error condition?

@debonair please add your findings here.

debonair:
@rcrichton please review

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -64,7 +64,7 @@
   "reports": {
     "enableReports": true,
     "dailyReportAt": "00 7 * * *",
-    "weeklyReportAt": "00 7 * * 6"
+    "weeklyReportAt": "00 7 * * 1"
   },
   "visualizer": {
     "enableVisualizer": true,
--- a/config/default.json
+++ b/config/default.json
@@ -62,7 +62,9 @@
     "consoleURL": "http://openhim.jembi.org"
   },
   "reports": {
-    "enableReports": true
+    "enableReports": true,
+    "dailyReportAt": "00 7 * * *",
+    "weeklyReportAt": "00 7 * * 6"
   },
   "visualizer": {
     "enableVisualizer": true,
--- a/config/default.json
+++ b/config/default.json
@@ -63,8 +63,8 @@
   },
   "reports": {
     "enableReports": true,
-    "dailyReportAt": "00 7 * * *",
-    "weeklyReportAt": "00 7 * * 1"
+    "dailyReportAt": "0 7 * * *",
+    "weeklyReportAt": "0 7 * * 1"
   },
   "visualizer": {
     "enableVisualizer": true,
--- a/src/reports.coffee
+++ b/src/reports.coffee
@@ -7,6 +7,8 @@ config.reports = config.get('reports')
 contact = require './contact'
 metrics = require './metrics'
 moment = require "moment"
+config = require "./config/config"
+config.reports = config.get('reports')
 
 # Function Sends the reports
 
@@ -94,12 +96,12 @@ sendUserEmail = (report) ->
 
 fetchChannelReport = (channel, user, flag, callback) ->
   if flag == 'dailyReport'
-    from = moment().startOf('day').toDate()
-    to = moment().endOf('day').toDate()
+    from = moment().subtract(1, 'days').startOf('day').toDate()
+    to = moment().subtract(1, 'days').endOf('day').toDate()
     period = 'day'
   else
-    from = moment().startOf('week').toDate()
-    to = moment().endOf('week').toDate()
+    from = moment().subtract(1, 'days').startOf('week').toDate()
+    to = moment().subtract(1, 'days').endOf('week').toDate()
     period = 'week'
 
   item = {}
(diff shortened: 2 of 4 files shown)


rcrichton:
Hey @debonair I can't comment on individual lines as some of the problem aren't in the changes you made, so I'l just mention them here and link to the lines:

* When the data array comes back empty the report templates fail, we need to add some null checks for these. Also, that `do (data)` isn't needed. See [here](https://github.com/jembi/openhim-core-js/pull/484/files?diff=unified#diff-28a2f278c5140aaf68de20b200ceac17R135) and [here](https://github.com/jembi/openhim-core-js/pull/484/files?diff=unified#diff-28a2f278c5140aaf68de20b200ceac17R172)
* I think we need to be a bit more specific with the agenda times, as it is now if a report gets run in the middle of a week/day then we any get half a day/weeks data. Maybe we should set the daily ones to run at exactly 2am each day and report the previous days data and the weekly ones at 2am after the last day of the week and report the previous weeks data. See [here](https://github.com/jembi/openhim-core-js/pull/484/files?diff=unified#diff-28a2f278c5140aaf68de20b200ceac17R200) and [here](https://github.com/jembi/openhim-core-js/pull/484/files?diff=unified#diff-28a2f278c5140aaf68de20b200ceac17R96)
* Somehow the exceptions are getting swallowed when reports fail. I'm not sure where these are going, perhaps agenda is swallowing them. Perhaps look into this, so we can log errors.

rcrichton:
It looks like we can detect errors or success using job events. Maybe we should add some event handler with some logging so we can more easily see what is going on? https://github.com/rschmukler/agenda#job-queue-events

debonair:
@rcrichton I think  the reports should default to run at 7 am showing the previous day's data so that there is no ambiguity in time zone.  Also the times  should be specified in the config file so that the setting can be tailored to a particular implementation.

rcrichton:
@debonair sure, that sounds good. A configurable time would be best.

debonair:
@rcrichton you may review, just waiting for travis builds to pass

src/reports.coffee
```diff
@@ -131,19 +134,22 @@ plainTemplate = (report) ->
...
+                  Successful:  #{ data.statusData[0].successful }  \r\n
+                  Processing: #{ data.statusData[0].processing }  \r\n
+                  Completed:  #{ data.statusData[0].completed }  \r\n
+                  Completed with errors: #{ data.statusData[0].completedWErrors } \r\n \r\n
+                  <---------- End Channel -------------------------------------------------> \r\n \r\n
+                \r\n
+                \r\n
+              "
+      else
+        text += "There is no data in this"
```
rcrichton: Instead of writing there is not data, maybe, we should rather just display values of zero. You may be able to add the null check in the string expressions above. Like `#{ if data.data[0] then data.data[0].load else 0}`

src/reports.coffee
```diff
@@ -168,14 +174,17 @@ htmlTemplate = (report) ->
...
+        text += "<tr><td><i>#{data.channel.name}</i></td>"
+        text += "<td> #{ data.data[0].load } transactions </td>"
+        text += "<td> #{ data.data[0].avgResp } </td>"
+        text += "<td> #{ data.statusData[0].failed }  </td>"
+        text += "<td> #{ data.statusData[0].successful }  </td>"
+        text += "<td> #{ data.statusData[0].processing }  </td>"
+        text += "<td> #{ data.statusData[0].completed }  </td>"
+        text += "<td> #{ data.statusData[0].completedWErrors } </td></tr>"
+      else
+        text += "<tr><td>There is no data in this report</td></tr>"
```
rcrichton: Instead of writing there is not data, maybe, we should rather just display values of zero. You may be able to add the null check in the string expressions above. Like `#{ if data.data[0] then data.data[0].load else 0}`

src/reports.coffee
```diff
@@ -196,8 +205,14 @@ setupAgenda = (agenda) ->
...
-  agenda.every "1 weeks", 'send weekly channel metrics'
-  agenda.every "1 days", 'send daily channel metrics'
+  agenda.every config.reports.weeklyReportAt, 'send weekly channel metrics'
+  agenda.every config.reports.dailyReportAt, 'send daily channel metrics'
+
+  agenda.on "start", (job)->
+    logger.info "starting job: " + job.attrs.name
+
+  agenda.on "fail", (err, job)->
+    logger.error "Job " + job.attrs.name + " failed with " + err.message
```
rcrichton: Perhaps add an event listener for 'complete' tasks as well.

Also, perhaps we should setup these event [here rather](https://github.com/jembi/openhim-core-js/blob/master/src/server.coffee#L202-L204) as places other than reports use agenda, so it will make it a bit clearer that jobs will be logged.

src/reports.coffee
```diff
@@ -50,9 +52,11 @@ sendReports = (job, flag, done) ->
 #     Pre-Fetch report data into Channel Map
       for key , obj of channelMap
         innerDeferred = Q.defer()
-        fetchChannelReport obj.channel, obj.user, flag, (item) ->
-          channelReportMap[key] = item
-          innerDeferred.resolve()
+        do (innerDeferred,key, obj) ->
```
rcrichton: Minor style comment: you may want a space before 'key'

config/default.json
```diff
@@ -62,7 +62,9 @@
     "consoleURL": "http://openhim.jembi.org"
   },
   "reports": {
-    "enableReports": true
+    "enableReports": true,
+    "dailyReportAt": "00 7 * * *",
+    "weeklyReportAt": "00 7 * * 1"
```
rcrichton: I think its more normal to just use a 0 instead of 00.

rcrichton:
Hey @debonair this seems to be working for me. I've added a few more comments above.

debonair:
@rcrichton please review

src/server.coffee
```diff
@@ -202,6 +202,16 @@ else
...
         address: config.mongo.url
+
+    agenda.on "start", (job)->
+      logger.info "starting job: " + job.attrs.name
+
+    agenda.on "fail", (err, job)->
+      logger.error "Job " + job.attrs.name + " failed with " + err.message
+
+    agenda.on "complete", (job)->
+      logger.error "Job " + job.attrs.name + " has completed"
```
rcrichton: This should used logger.info not error.

src/reports.coffee
```diff
@@ -169,13 +172,13 @@ htmlTemplate = (report) ->
...
-      text += "<td> #{ data.statusData[0].processing }  </td>"
-      text += "<td> #{ data.statusData[0].completed }  </td>"
-      text += "<td> #{ data.statusData[0].completedWErrors } </td></tr>"
+      text += "<td> #{ if data.data[0].load then data.data[0].load else 0 } transactions </td>"
+      text += "<td> #{ if data.data[0].avgResp then data.data[0].avgResp else 0 } </td>"
+      text += "<td> #{ if data.statusData[0].failed then data.statusData[0].failed else 0 }  </td>"
+      text += "<td> #{ if data.statusData[0].successful then data.statusData[0].successful else 0 }  </td>"
+      text += "<td> #{ if data.statusData[0].processing then data.statusData[0].processing else 0 }  </td>"
+      text += "<td> #{ if data.statusData[0].completed then data.statusData[0].completed else 0 }  </td>"
+      text += "<td> #{ if data.statusData[0].completedWErrors then data.statusData[0].completedWErrors else 0 } </td></tr>"
```
rcrichton: @debonair this doesn't work, are you testing these before asking me to review it? This has the same problem that we were originally trying to solve. You null checks are too specific, you need to be testing data.data[0] and data.statusData[0] if they are null. You could use the existential operator if you want to check both. Something like: `data.data[0]?.load?` data.data[0] doesn't get checked.

rcrichton:
@debonair more comments. Also, it seems that when errors get thrown in the templates its not getting output by the agenda on fail event. Maybe you can look into this. That would make it easier for me to figure out where its failing.

debonair:
@rcrichton seems you need to specifically fail a job on error using job.fail('failure reason'); I am looking into this now

debonair:
@rcrichton I have tested and I think I have ironed out all the issues now. please take a look

rcrichton:
Thanks @debonair, this is now working well :+1:

PR #485, opened 2015-04-28
Added check to see if keystore exists

devcritter:
Hey @rcrichton ran into a problem while doing a fresh install (vagrant). Just added a check to see if the keystore exists.


rcrichton:
Thanks @devcritter good catch. :+1:

PR #486, opened 2015-04-28
Added git install dependency.

devcritter:
@rcrichton yet another quick ad-hoc PR :) This just adds git as a dependency for those npm packages that need it.


rcrichton:
:+1:

PR #489, opened 2015-04-30
Keystore validity return false if a protected key is tested without a pasphrase

rcrichton:
@debonair could you review this for me please?


debonair:
Just waiting for build to pass, otherwise, looks good +1 I have tested this with console

PR #491, opened 2015-05-05
Async non primary routes

debonair:
@rcrichton please review


src/middleware/router.coffee
```diff
@@ -84,6 +85,8 @@ handleServerError = (ctx, err) ->
 
 sendRequestToRoutes = (ctx, routes, next) ->
   promises = []
```
rcrichton: It seems that this doesn't need to be an array anymore if it is only dealing with a single primary promise. Maybe naming it primaryPromise would be better to.

src/middleware/router.coffee
```diff
@@ -134,13 +139,25 @@ sendRequestToRoutes = (ctx, routes, next) ->
...
+
 
       promises.push promise
+      nonPrimaryPromises.push nonPrimaryPromise
 
     (Q.all promises).then ->
       next()
 
+    (Q.all nonPrimaryPromises).then ->
+      messageStore.storeResponse ctx, ->
```
rcrichton: Perhaps we should split out the storeReponse function into a storePrimaryResponse function and a storeNonPrimaryResponse function. Then for non primary responses we can just call that and it only adds to the routes array instead of updating the whole message again. Also, this will prevent the transaction status from changing as new routes get completed.

For the transaction status what we actually want is for it to be set to remain as 'Processing' and then only be updated finally once all routes have returned.

src/middleware/router.coffee
```diff
@@ -134,13 +139,25 @@ sendRequestToRoutes = (ctx, routes, next) ->
...
+        logger.info 'Storing non primary route responses'
+        stats.incrementTransactionCount ctx, ->
+        stats.measureTransactionDuration ctx, ->
```
rcrichton: We may also want to split out the stats function into primary and nonPrimary functions, similar to above.

src/middleware/router.coffee
```diff
@@ -166,6 +183,7 @@ buildNonPrimarySendRequestPromise = (ctx, route, options, path) ->
 
     ctx.routes = [] if not ctx.routes
     ctx.routes.push routeObj
```
rcrichton: If we make the changes I've mentioned above then we won't need to attach the routes to the context at all, we could just pass the routeObj directly to the relevent messageStore function.

src/middleware/router.coffee
```diff
@@ -166,6 +183,7 @@ buildNonPrimarySendRequestPromise = (ctx, route, options, path) ->
...
+    exports.nonPrimaryRoutes.push routeObj
```
rcrichton: I'm not quite sure what this is used for?

rcrichton:
Hey @debonair this looks pretty good so far. I've added some comment above that I think will improve the workflow.

src/middleware/router.coffee
debonair: I needed this to keep track of the async non primary routes for my tests. because once the response has been sent to the client, the context gets reset.

src/middleware/messageStore.coffee
```diff
@@ -65,70 +69,119 @@ exports.storeTransaction = (ctx, done) ->
     else
       ctx.transactionId = tx._id
       ctx.header['X-OpenHIM-TransactionID'] = tx._id.toString()
+      ctx.header['channel-id'] = tx.channelID
       return done null, tx
 
 exports.storeResponse = (ctx, done) ->
   logger.info 'Storing response for transaction: ' + ctx.transactionId
+  getNumRoutes ctx.authorisedChannel._id, (numRoutes) ->
```
rcrichton: The number of route returned here doesn't seem to be used.

src/middleware/messageStore.coffee
```diff
@@ -65,70 +69,119 @@ exports.storeTransaction = (ctx, done) ->
...
+      timestamp: ctx.response.timestamp
+
+
+    # check if channel response body is false and remove
+    if ctx.authorisedChannel.responseBody == false
+      # reset request body - primary route
+      res.body = ''
+
+      # reset request body - routes
+      if ctx.routes
```
rcrichton: We don't need to do this route checking if we aren't storing route responses here. Just make sure this logic moves to where we are storing route responses.

src/middleware/router.coffee
rcrichton: This doesn't seem to be used anymore.

rcrichton: Nevermind, I see where it is used. But it doesn't need to be an array still.

src/middleware/router.coffee
```diff
@@ -91,56 +94,73 @@ sendRequestToRoutes = (ctx, routes, next) ->
...
+              messageStore.storeNonPrimaryResponse ctx, response, ->
+                stats.incrementTransactionCount ctx, ->
+                stats.measureTransactionDuration ctx, ->
+            catch err
+              logger.error err
+
+        nonPrimaryPromises.push nonPrimaryPromise
+
+    (Q.all nonPrimaryPromises).then ->
+      logger.info "all non primary routes completed"
```
rcrichton: This may be a better place to do the final transaction status update. Perhaps, create a new function in the message store that does the final update and call it here.

src/middleware/router.coffee
```diff
@@ -91,56 +94,73 @@ sendRequestToRoutes = (ctx, routes, next) ->
...
+
+
+exports.nonPrimaryRoutes = []
```
rcrichton: I think you can get rid of this.

src/middleware/router.coffee
rcrichton: I think this comment above still applies.

src/middleware/messageStore.coffee
```diff
@@ -65,70 +69,119 @@ exports.storeTransaction = (ctx, done) ->
...
+    getNumRoutes ctx.request.header["channel-id"], (numRoutes) ->
+    #  Get the current transaction and get the number of routes that have completed
+      do (numRoutes) ->
+        if ctx.request.header?["X-OpenHIM-TransactionID"]?
+          transactions.Transaction.findById ctx.request.header["X-OpenHIM-TransactionID"], (err,tx) ->
+            do (tx) ->
+              numRouteResps = tx.routes.length + 1
+              remainingRoutes = numRoutes - numRouteResps
+              logger.info "num remaining routes: #{remainingRoutes}"
+              isLastRoute = false
```
rcrichton: Rather do the last route processing in another function and call it when the non primary route promises resolve. You have this already in the router.

rcrichton:
Hey @debonair its looking good. I've added a few comments above to think about. Also, you will need to think about adding a test or two that ensures that the final transaction status is only updated at the end after all routes are done. But we are getting closer.

debonair:
@rcrichton please review

src/middleware/messageStore.coffee
```diff
@@ -7,6 +7,8 @@ statsdServer = config.get 'statsd'
 application = config.get 'application'
 SDC = require 'statsd-client'
 os = require 'os'
+co = require "co"
```
rcrichton: I don't think this is used.

src/middleware/messageStore.coffee
```diff
@@ -7,6 +7,8 @@ statsdServer = config.get 'statsd'
...
+Channel = require('../model/channels').Channel
```
rcrichton: Or this, unless I'm missing something.

src/middleware/messageStore.coffee
```diff
@@ -18,6 +20,8 @@ exports.transactionStatus = transactionStatus =
   COMPLETED_W_ERR: 'Completed with error(s)'
   FAILED: 'Failed'
 
+exports.routeStatus = routeStatus = true
```
rcrichton: Not sure where this is used either.

src/middleware/messageStore.coffee
```diff
@@ -105,31 +88,72 @@ exports.storeResponse = (ctx, done) ->
...
   if ctx.mediatorResponse
     update.orchestrations = ctx.mediatorResponse.orchestrations if ctx.mediatorResponse.orchestrations
     update.properties = ctx.mediatorResponse.properties if ctx.mediatorResponse.properties
 
-  transactions.Transaction.findOneAndUpdate { _id: ctx.transactionId }, update, (err, tx) ->
+  transactions.Transaction.findOneAndUpdate { _id: ctx.transactionId }, update , { runValidators: true }, (err, tx) ->
+    logger.info "stored primary response for #{tx._id}"
     if err
-      logger.error 'Could not save response metadata for transaction: ' + ctx.transactionId + '. ' + err
+      logger.info 'Could not save response metadata for transaction: ' + ctx.transactionId + '. ' + err
```
rcrichton: Perhaps we should change this log message back to error.

src/middleware/router.coffee
rcrichton: @debonair do we still need to attach the `routeObj` to the `ctx` with this async processing. Maybe we can rather just use the route object directly. I think when you return the `routeObj` it will become available as in the `.then` function here: https://github.com/jembi/openhim-core-js/pull/491/files#diff-ffc48656de2a6c6b5ee8a785b33109b0R153

Chat to me more about this if you like, I'm if what I've said above is exactly how it works, but I think it is.

src/middleware/messageStore.coffee
```diff
@@ -105,31 +88,72 @@ exports.storeResponse = (ctx, done) ->
...
       return done err
     return done()
 
+exports.storeNonPrimaryResponse = (ctx, done) ->
+  # check if channel response body is false and remove
+  if ctx.authorisedChannel.responseBody == false
+    response.response.body = ''
+
+  if ctx.transactionId?
+    transactions.Transaction.findByIdAndUpdate ctx.transactionId, { routes: ctx.routes }, (err,tx) ->
```
rcrichton: This seems to save all of the routes each time its called. I think we would rather want to save only one route at a time instead of all of them.

test/integration/e2eIntegrationTests.coffee
```diff
@@ -861,13 +862,14 @@ describe "e2e Integration Tests", ->
...
                 done err
               else
-                Transaction.findOne {}, (err, res) ->
-                  res.status.should.be.equal mediatorResponse.status
-                  res.orchestrations.length.should.be.exactly 1
-                  res.orchestrations[0].name.should.be.equal mediatorResponse.orchestrations[0].name
-                  should.exist res.properties
-                  res.properties.orderId.should.be.equal mediatorResponse.properties.orderId
-                  done()
+                Transaction.findOne {}, (err1, res1) ->
```
rcrichton: What is the purpose of this extra query?

test/integration/tcpIntegrationTests.coffee
```diff
@@ -201,11 +201,12 @@ describe "TCP/TLS/MLLP Integration Tests", ->
         data.should.be.exactly 'TLS OK'
         done()
 
-  it "should return an error when the client cert is not known by the server", (done) ->
-    server.start tcpHttpReceiverPort: 7787, ->
-      sendTCPTestMessage 4004, (data) ->
-        data.should.be.exactly 'An internal server error occurred'
-        done()
+#  it "should return an error when the client cert is not known by the server", (done) ->
```
rcrichton: This needs to be uncommented.

test/integration/e2eIntegrationTests.coffee
debonair: I could not find a cleaner way to delay checking for the updated status, the test was checking for the status before it had been updated,.

test/unit/routerTest.coffee
```diff
@@ -203,19 +217,20 @@ describe "HTTP Router", ->
...
+                ctx.routes[0].response.headers.should.be.ok
+                ctx.routes[0].request.path.should.be.exactly "/test/multicasting"
+                ctx.routes[0].request.timestamp.should.be.exactly requestTimestamp
+                ctx.routes[1].response.status.should.be.exactly 400
+                ctx.routes[1].response.body.toString().should.be.eql "Mock response body 3\n"
+                ctx.routes[1].response.headers.should.be.ok
+                ctx.routes[1].request.path.should.be.exactly "/test/multicasting"
+                ctx.routes[1].request.timestamp.should.be.exactly requestTimestamp
+                done()
+              ), 5000
```
rcrichton: We probably only need a few 100ms here. Just to make the tests run quicker.

test/integration/e2eIntegrationTests.coffee
rcrichton: Ok, I see. It may be better to rather set a very small setTimeout.

src/middleware/messageStore.coffee
debonair: I had a lot of issues with this. for whatever reason, i could not push new route objects in to the transaction document. the document would end up with only the last route pushed. So I resorted to pushing them to the context rather

rcrichton:
Hey @debonair, a few comment above. Glad all the test are working now.

debonair:
@rcrichton please review

config/default.json
```diff
@@ -73,8 +73,8 @@
     "orchestrationTSBufferMillis": 100
   },
   "statsd": {
-    "enabled": false,
-    "host": "127.0.0.1"
+    "enabled": true,
+    "host": "104.236.15.32"
```
rcrichton: Can you change these back to the previous defaults.

src/middleware/messageStore.coffee
```diff
@@ -65,32 +65,11 @@ exports.storeTransaction = (ctx, done) ->
     else
       ctx.transactionId = tx._id
       ctx.header['X-OpenHIM-TransactionID'] = tx._id.toString()
+      ctx.header['channel-id'] = tx.channelID
```
rcrichton: I can't see where this is used, perhaps it can be removed?

src/middleware/messageStore.coffee
```diff
@@ -130,6 +100,56 @@ exports.storeResponse = (ctx, done) ->
       return done err
     return done()
 
+exports.storeNonPrimaryResponse = (ctx, routeObject, done) ->
+  # check if channel response body is false and remove
+  if ctx.authorisedChannel.responseBody == false
+    response.response.body = ''
+
+  if ctx.transactionId?
```
rcrichton: Perhaps add an else statement here and log and error if the transactionId doesn't exist.

src/middleware/router.coffee
```diff
@@ -91,55 +98,74 @@ sendRequestToRoutes = (ctx, routes, next) ->
...
+                  stats.nonPrimaryRouteDurations ctx, routeObj, ->
+                    console.log 'sent durarions'
+            catch err
+              logger.error err
+
+        promises.push promise
     (Q.all promises).then ->
-      next()
+      messageStore.setFinalStatus ctx, ->
+        logger.info "all routes completed"
```
rcrichton: Maybe add a little more info to this log message, like the transaction id it relates to.

src/middleware/router.coffee
```diff
@@ -166,6 +192,7 @@ buildNonPrimarySendRequestPromise = (ctx, route, options, path) ->
 
     ctx.routes = [] if not ctx.routes
```
rcrichton: Can we delete this `ctx.route` variable now or is it used elsewhere?

src/middleware/router.coffee
```diff
@@ -91,55 +98,74 @@ sendRequestToRoutes = (ctx, routes, next) ->
...
+          logger.info "executing non primary: #{route.name}"
+          promise = buildNonPrimarySendRequestPromise(ctx, route, options, path)
+          .then (routeObj) ->
+            logger.info "Storing non primary route responses #{route.name}"
+
+            try
+              messageStore.storeNonPrimaryResponse ctx, routeObj, ->
+                stats.nonPrimaryRouteRequestCount ctx, routeObj, ->
+                  stats.nonPrimaryRouteDurations ctx, routeObj, ->
+                    console.log 'sent durarions'
```
rcrichton: Remove this console.log.

src/middleware/router.coffee
debonair: there are some router tests that rely on this, there is no easy way to check if the router can route to multiple routes without checking the database, and doing that would replicate an end to end test

debonair: So I kept pushing the route responses to the context

rcrichton: Ok, I think that is ok in that case.

rcrichton:
Hey @debonair this looks good to me. Just a few minor comments above. Let me know once those are done and I'l have another quick look.

debonair:
@rcrichton they are done please take another look

src/middleware/stats.coffee
```diff
@@ -20,52 +20,8 @@ exports.incrementTransactionCount = (ctx, done) ->
     sdc.increment domain + '.channels.' + ctx.authorisedChannel._id + '.statuses.' + transactionStatus # Per Channel Status
 
     #Collect stats for non-primary routes
```
rcrichton: Can you remove these next 3 commented out lines.

src/middleware/stats.coffee
```diff
@@ -130,25 +86,8 @@ exports.measureTransactionDuration = (ctx, done) ->
     sdc.timing domain + '.channels.' + ctx.authorisedChannel._id + '.statuses.' + transactionStatus, ctx.timer # Per Channel Status
 
     #Collect stats for non-primary routes
```
rcrichton: And these next 3 as well.

rcrichton:
@debonair cool, just noticed two more minor things. After those are done then :+1:

PR #503, opened 2015-05-26
Mail templates

debonair:
@rcrichton please review


src/reports.coffee
```diff
@@ -84,7 +87,35 @@ sendReports = (job, flag, done) ->
             report.type = 'Daily'
           else
             report.type = 'Weekly'
+
+          report.instance = config.alerts.himInstance
+
           try
+            i = 0
+            for data in report.data
```
rcrichton: Minor: I think here you could use `for data, i in report.data` then you won't have to have your own `i` variable or increment it.

src/templates/daily_report/html.handlebars
```diff
@@ -0,0 +1,517 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
rcrichton: This is used for both daily and weekly reports and it also its coffeescript, so maybe move out of src/ and rename it to `templates/report.handlebars`

src/reports.coffee
```diff
@@ -152,44 +185,13 @@ plainTemplate = (report) ->
...
-      text += "<td> #{ if data.statusData[0]?.completedWErrors? then data.statusData[0].completedWErrors else 0 } </td></tr>"
-  text += "
-    </table>
-    </div>
-    </body>
-    </html>
-    "
-  text
+renderTemplate = (templateName, templateData, callback) ->
+  templateDir = path.join(__dirname, '../src/templates', templateName)
```
rcrichton: There are some issues with linking to files when the openhim is used as an npm package, so it would be best to use our appRoot global instead of `__dirname` for consistency: https://github.com/jembi/openhim-core-js/blob/master/src/server.coffee#L6

package.json
```diff
@@ -16,9 +16,11 @@
     "bcryptjs": "^2.0.0",
     "co": "^4.1.0",
     "cookie": "^0.1.2",
+    "email-templates": "^2.0.0-beta",
```
rcrichton: I'm a little worried about using a beta, but I think this is ok until there is a release as we aren't using this much. Maybe create and issue to switch us to a release when its ready.

src/reports.coffee
```diff
@@ -84,7 +87,35 @@ sendReports = (job, flag, done) ->
...
+                report.data[i].completed = (if data.statusData[0]?.completed? then data.statusData[0].completed else 0)
+                report.data[i].completedWErrors = (if data.statusData[0]?.completedWErrors? then data.statusData[0].completedWErrors else 0)
+                report.data[i].loadStyle = (if report.data[i].load > 0 then '' else colorGrey)
+                report.data[i].avgRespStyle = (if report.data[i].avgResp > 0 then '' else colorGrey)
+                report.data[i].failedStyle = (if report.data[i].failed > 0 then 'color: red;' else colorGrey)
+                report.data[i].successfulStyle = (if report.data[i].successful > 0 then '' else colorGrey)
+                report.data[i].processingStyle = (if report.data[i].processing > 0 then '' else colorGrey)
+                report.data[i].completedStyle = (if report.data[i].completed > 0 then 'color: orange;' else colorGrey)
+                report.data[i].completedWErrorsStyle = (if report.data[i].completedWErrors > 0 then 'color: orangered;' else colorGrey)
+                report.data[i].rowColor = rowColor
```
rcrichton: It would be better to keep all display stuff out of the data as its good practice, is it not possible to do this with pure css. If its hard then I'm ok with this way.

rcrichton: It seems its possible in pure css: http://stackoverflow.com/questions/3084261/alternate-table-row-color-using-css

debonair: I tried this, it does not work, most email clients support only inlined styles

src/templates/daily_report/html.handlebars
```diff
@@ -0,0 +1,517 @@
...
+<html>
+<head>
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+
+    <!-- Facebook sharing information tags -->
+    <meta property="og:title" content="*|MC:SUBJECT|*" />
+
+    <title>*|MC:SUBJECT|*</title>
+    <style type="text/css">
+        /* Client-specific Styles */
```
rcrichton: It seems there is a lot of css in here, it may be good to remove any css that we aren't using.

src/templates/daily_report/html.handlebars
```diff
@@ -0,0 +1,517 @@
...
+                                        <table border="0" cellpadding="20" cellspacing="0" width="100%">
+                                            <tr>
+                                                <td valign="top" align="center" class="bodyContent">
+                                                    <div mc:edit="std_content00">
+                                                        <h1 class="h1" style="color: darkgreen">{{type}} Transaction  Report for {{instance}}</h1>
+                                                    </div>
+                                                </td>
+                                            </tr>
+                                            <tr>
+                                                <td valign="top" style="padding-top:0; padding-bottom:0;">
```
rcrichton: I think it would be better to put these styles in to css classes rather as they repeat a lot. You are already using a css script above.

rcrichton:
Hey @debonair, I've added some comments above.

src/reports.coffee
rcrichton: But it seems we are using an inline css script, could we not put it there? https://github.com/jembi/openhim-core-js/pull/503/files#diff-5778af25af0ff053ad8fee2d01e80aa4R10

src/reports.coffee
debonair: @rcrichton  I am not sure how to use that... I see where it is defined but not how is it used, please paste here,  what line 189 should be

rcrichton: appRoot is the root directory of the openhim project. So you can link to the file from the root directory. See its other uses: https://github.com/jembi/openhim-core-js/search?utf8=%E2%9C%93&q=appRoot

src/templates/daily_report/html.handlebars
debonair: non of that css was being used so I removed it. Only the inline styles work

src/templates/daily_report/html.handlebars
debonair: Only inline css styles work with clients like gmail

debonair:
@rcrichton please review. I have fixed and commented on the issues raised

rcrichton:
Ok, thanks @debonair looks good :+1: Seems much cleaner without all the extra CSS.

Issue #502, opened 2015-05-26
Mediators are allowed to override the status string


devcritter:
A mediator can optionally add a `status` field to their response which should set the status string for a transaction. It seems this is currently not working.

devcritter:
Closed via #504

Issue #501, opened 2015-05-26
Default handling for 4xx mediator responses


devcritter:
When a mediator responds with a 4xx response, transactions currently get left in the 'Processing' status:
> 2015-05-26T08:34:37.738Z - info: [worker1] primary route completed
2015-05-26T08:34:37.741Z - info: [worker1] Storing visualizer events for transaction: 5564301d7413f3cc07689e29
2015-05-26T08:34:37.742Z - info: [worker1] stored primary response for 5564301d7413f3cc07689e29
2015-05-26T08:34:37.743Z - info: [worker1] Final status for transaction 5564301d7413f3cc07689e29 : Processing

For reference - see a summary of the status string algorithm [here](https://github.com/jembi/openhim-core-js/issues/93)

devcritter:
Closed via #504

PR #504, opened 2015-05-27
fixed issues with status closes #500 #501 #502

### LINKED ISSUE #500: Bug: Core crashing
I found the below bug in core when sending through a transaction: Core crashes when it doesnt find response.timestamp. Could potential be due to a mediator error being used but core error should be handled
screenshot from 2015-05-25 16 23 18

debonair:
@devcritter please review

### DIFF
--- a/src/middleware/messageStore.coffee
+++ b/src/middleware/messageStore.coffee
@@ -85,6 +85,10 @@ exports.storeResponse = (ctx, done) ->
 
   update = { response: res }
 
+  # Set status from mediator
+  if ctx.mediatorResponse?.status?
+    update.status = ctx.mediatorResponse.status
+
   if ctx.mediatorResponse
     update.orchestrations = ctx.mediatorResponse.orchestrations if ctx.mediatorResponse.orchestrations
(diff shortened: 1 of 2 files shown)


src/middleware/messageStore.coffee
```diff
@@ -122,32 +126,38 @@ exports.setFinalStatus = setFinalStatus = (ctx, callback) ->
...
-    # In all other cases mark as completed
-    if ctx.status is null or ctx.status is undefined
-      tx.status = transactionStatus.COMPLETED
-
-    logger.info "Final status for transaction #{tx._id} : #{tx.status}"
-    transactions.Transaction.findByIdAndUpdate transactionId, {status: tx.status}, { },  (err,tx) ->
-      tx.save
-      callback tx
+      # In all other cases mark as completed
+      if ctx.status is null or ctx.status is undefined
```
devcritter: Could probably just say: `if not ctx.status`

devcritter:
Thanks @debonair! Looks good. Just one very minor comment but otherwise :+1: I tested against the mediator and solves the issues.

Issue #505, opened 2015-05-28
Store primary route request and response separately


rcrichton:
Currently we store the request and response in a transaction, however this is currently representing how the request and response are seen by the client and we don't have a way to store the request and response that we make for the primary route. These are often mostly the same, however paths could be altered and the host and port that we are sending to will obviously differ.

Should we store a route in the transaction.routes array for the primary route as well? Will this be too much duplication?

PR #506, opened 2015-05-28
Adds host and port metadata for routes.

### LINKED ISSUE #497: Add server to request model
We currently only capture the `path` for a [request](https://github.com/jembi/openhim-core-js/blob/d6bea4c94dbcfc4b84ae625160c6d33f1fb65c35/src/model/transactions.coffee#L7-L13). It would be useful to capture the server information as well: specifically the **host** and **port** that the request got sent to.

This would be good for orchestrations as the information isn't captured at all (for the main request this information is implicit in the channel route).

rcrichton:
Closes #497

@devcritter could you have a look at this? Also, while working on this I realised something, see #505.


devcritter:
:+1: Looks good

PR #507, opened 2015-06-01
Adds url rewriting middleware

### LINKED ISSUE #262: URLs returned back in a response body (or headers) need to point to the HIM not the service consumer
(URL rewritting)

rcrichton:
Closes #262 

@devcritter could you please review this for me.

When enabled it rewrites the responses based the currently configured channel or user specified configuration.


src/model/channels.coffee
```diff
@@ -23,6 +23,13 @@ AlertsDef =
   "users":        [ContactUserDef]
   "failureRate":  Number
 
+RewriteRuleDef =
+  "fromHost":       type: String, required: true
+  "toHost":         type: String, required: true
+  "fromPort":       type: Number, require: true, default: 80
```
devcritter: Shouldn't it be required with a -d?

src/middleware/rewriteUrls.coffee
```diff
@@ -0,0 +1,104 @@
...
+
+    callback null, rwConfig
+
+rewriteUrls = (body, channel, authType, callback) ->
+  exports.fetchRewriteConfig channel, authType, (err, rwConfig) ->
+    if err?
+      return callback err
+
+    # rewrite each found href attribute (in JSON or XML)
+    newBody = body.replace /["|']?href["|']?[:|=]\s?["|'](\S*?)["|']/g, (match, hrefUrl) ->
```
devcritter: Maybe we should check for `src=` as well? Just looking at a CSD page for reference. In fact it might be good just to test against CSD to see that this works for it, because, I believe, this issue was largely motivated by it?

src/middleware/rewriteUrls.coffee
```diff
@@ -0,0 +1,104 @@
...
+  utils.getAllChannels (err, channels) ->
+    if err?
+      return callback err
+
+    for channel in channels
+      for route in channel.routes
+        do ->
+          if route.primary
+            if route.pathTransform
+              inverveTransform = invertPathTransform route.pathTransform
```
devcritter: type: inverve

src/middleware/rewriteUrls.coffee
```diff
@@ -0,0 +1,104 @@
...
+  rwConfig = []
+  if channel.rewriteUrlsConfig?
+    rwConfig = channel.rewriteUrlsConfig
+
+  # add in default rewrite rules for hosts we proxy
+  utils.getAllChannels (err, channels) ->
+    if err?
+      return callback err
+
+    for channel in channels
```
devcritter: So does this mean rewrite rules for other channels may get applied? Perhaps we should keep it to just the current channel? - it'll avoid any unexpected rewrites. Plus, although probably not a security risk, perhaps it's best if a transaction doesn't trigger calls that touch other channels?

src/middleware/rewriteUrls.coffee
```diff
@@ -0,0 +1,104 @@
...
+Q = require 'q'
+
+Channel = (require '../../lib/model/channels').Channel
+config = require '../config/config'
+routerConf = config.get 'router'
+utils = require '../../lib/utils'
+router = require '../../lib/middleware/router'
+
+invertPathTransform = (pathTransform) ->
+  return pathTransform.replace /s\/(.*?)\/(.*?)($|\/(.*)$)/, 's/$2/$1/$4'
```
devcritter: To be honest inverting the transform confused me a bit. Esp. when reading the tests. The transform is definitely useful but maybe it would be clearer "un-inverted"? (verted? :) )

I think I understand the reasoning - since it's `s/from/to/` it helps to have `to` last since it's the `to` server - but maybe we should think about the transform differently? Perhaps we should rather think of it as `s/regex/replacement/`?

devcritter:
@rcrichton some comments above, but in general it looks good! nice one!

src/model/channels.coffee
rcrichton: It should...

src/middleware/rewriteUrls.coffee
rcrichton: Good idea, the request that I was modelling this off was from Justin to support links in FHIR. That is covered by `href=`. I'l add support for `src=` as well and add a test with a sample CSD doc. I wonder if we should make this regex configurable or if that just gets too complex?

src/middleware/rewriteUrls.coffee
rcrichton: Doh

src/middleware/rewriteUrls.coffee
rcrichton: Well not really, what it does is finds which hosts we proxy that are on primary routes, then it creates 'virtual' rules for them. So for example if the SHR for some reason sent back a link to a patient in the CR (using a direct link to the CR), then if we have a channel that points to the CR on a primary route we are able to rewrite the link to point to us instead because we know that host. It doesn't add the static, user configurable config from other channel, those are only applied per channel only, but it does create defaults for other  routes that we proxy. Does that make sense, if kinda gets a bit complex?

Perhaps we should have a config option on the channel for this, something like `autoRewriteUrlsToOtherChannels`?

:bulb: Ah, wait, looking at the code what I just explained isn't how it actually works, this is a bug, it shouldn't include user config for all channels, the the current one. You are right. Good catch.

rcrichton: Actually, sorry, I think its right. So I only add the static config on line 18 for the current channel, then iterate through all channel and add in the 'virtual' config for hosts we proxy.

src/middleware/rewriteUrls.coffee
rcrichton: Yeah this is confusing, so the inverting of the transform is only for 'virtual' config where we are rewriting a url for hosts we are proxying. In that case they may already have a pathTransform on the route that alters the path that they send a request to. So, when we get a link back from them in the response we need to do the reverse of that transform so that the link reflect the original path that the HIM channel knows about.

For example, say we have a channel with `urlPattern=/CSD/` and a pathTransform on the primary route as follows `pathTransform=s/CSD/ihris/` (ie. the actual server we are proxying is running on `http://<host>:<port>/ihirs/`). If we get links back from this server it will be something like `http://<host>:<port>/ihirs/something/123` but we need it to be `http://<him_host>:<him_port>/CSD/something/123`. To do this we can reverse (invert?) the pathTransform on the route (s/ihris/CSD/) and apply it while doing the rewrite (confusing I know).

This doesn't always works (eg. `s/CSD//`) but in that case a user can create a static rule for rewriting that will overwrite the 'virtual' rule because it is located first in the array (we only apply the first rule found).

I think I need to make the code a bit more readable, I'l add in some comments to help make this clear. Its pretty convoluted right now.

src/middleware/rewriteUrls.coffee
devcritter: Yeah, regex configurable could be useful, but yeah maybe in order to avoid the complexity for now, perhaps it's best to just go with href and src and open up another ticket?

src/middleware/rewriteUrls.coffee
devcritter: Aah okay; yeah I get it now. Apologies, just needed to get my head around this. Cool yeah, this looks good then.

Wrt adding a channel config option - I would say let's not bother with it until it becomes a problem.

src/middleware/rewriteUrls.coffee
devcritter: Right, ok cool thanks, this makes much more sense to me now.

devcritter:
@rcrichton thanks the explanations helped. Yeah it might good just add some explanatory comments (maybe even copy/paste your explanations from here?), but all looks good to me! :+1:

rcrichton:
Hey @devcritter, I've added some better comments into the source, added support for `src=` and added  in that channel config option to enable or disable 'virtual' rewrite rules (it default to true as I think this is generally what people would want).

Could you have a another quick look and let me know if all is ok?

rcrichton:
@devcritter any other comments here?

rcrichton:
I'm also not sure why the test are failing, it doesn't seem to be related to my code and works on my machine. I may have to look into that.

devcritter:
Apologies, didn't see the updates come through. Nope, looks good to me! :+1:

rcrichton:
Not sure why the random test are failing on travis, my machine works fine. I'm going to merge this in and write and issue to look into the tests.

PR #516, opened 2015-06-24
Issue514

### LINKED ISSUE #514: Transaction retrieve permissions aren't working
When logging into the console as a non-admin user, the transaction list is populated with transactions that they don't have permission to view. So it looks likely that the GET transactions endpoint isn't filtering correctly any more based on the user.

devcritter:
Closes #514 

@rcrichton if you could please take a look? thanks


rcrichton:
Thanks @devcritter, this looks good to me :+1:

PR #525, opened 2015-08-24
Updated vagrantfile to v2 and bumped the box mem.

devcritter:
Fixed the puppet mongodb module version to 0.10.0.

@rcrichton if you could please take a look? just some vagrant updates.

### DIFF
--- a/infrastructure/development/env/Vagrantfile
+++ b/infrastructure/development/env/Vagrantfile
@@ -1,28 +1,39 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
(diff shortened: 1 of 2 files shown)


devcritter:
The build keeps breaking but I can't see how it could have anything to do with this PR

rcrichton:
This looks good to me :+1:  I don't know why the build is breaking :( I'm seeing the same on other branches

devcritter:
Yeah, weirdly it seems to have passed on iojs. But even repeated attempts on node keeps breaking. Doesn't look like one our "normal" random breaks :/

PR #528, opened 2015-08-26
Audit server bug

BMartinos:
When send a audit message via TCP or TLS and no length is supplied then the request fails silently.

credit to @bausmeier for tracking down the issue


BMartinos:
Hey @rcrichton mind just doing a review on this fix. Thanks

bausmeier:
LGTM

rcrichton:
:+1: The tests seem to be failing but that seems to be happening on all branches atm :(

PR #529, opened 2015-08-26
Debian packaging

rcrichton:
@devcritter could you have a look over this for me?


packaging/create-deb.sh
```diff
@@ -0,0 +1,102 @@
...
+
+cd $HOME/targets
+TARGETS=(*)
+echo "Targets: $TARGETS"
+cd $HOME
+
+PKG=openhim-core-js
+
+echo -n "Which version of the OpenHIM (from NPM) would you like this package to install? (eg. 1.3.0) "
+read OPENHIM_VERSION
```
devcritter: Maybe add a check in case it's left blank (or maybe grep out the current version in package.json as a default value?)

devcritter:
@rcrichton I'm beginning to wonder if we shouldn't think of trying to include core precompiled in the package? Rather than have the package simply run npm install?

I'm busy testing the package atm in a clean vm, but it's taking ages to run and as you know, compiling those native deps takes a huge amount of mem. That said I'm not 100% certain how well it would work [including `node_modules/` in the dist], but might work. The idea I guess would be that the packaging script would need to npm install and build the native modules in `node_modules` for the architecture that the ubuntu package is targeting. (maybe it'd be enough to run npm install on a 64bit ubuntu 14.04 machine (ideal with vagrant) and including it in the package...). Or maybe use something like [this](https://github.com/jembi/openhim-core-js/wiki/Pre-Packaging-an-Offline-Release).

This would definitely add a lot of value to the packages and give it a separate use in cases where npm install is difficult (e.g. a box with low mem).

This would also provide a mechanism regarding testing release wrt @litlfred's [comment](https://github.com/jembi/openhim-console/pull/233#issuecomment-135400174) in the other PR.

devcritter:
FYI the package ran successfully on a clean ubuntu instance :)

rcrichton:
Hey Hannes, yeah I know what you mean. I just wasn't sure how the native modules needed to be built. I was worried that this wouldn't work down the line if you bundle everything in. Perhaps that offline-npm would work, but it just seems more complex to compile everything before hand and require a correct environment to build in when NPM is designed to do this. Also, this package should work for any version of ubuntu if we wanted to support multiple in the future.

That said, I had the same concerns as you when doing this. The install does take really long and the "package" is really just a script.

rcrichton:
Hey @devcritter, I've changed this to include all files for the openhim-core and its dependencies, even compiled native modules. Please let me know if this is better?

packaging/README.md
```diff
@@ -0,0 +1,6 @@
+Debian Packaging for the OpenHIM-core
+=====================================
+
+To create a debian package execute `./create-deb.sh`. This will run you through the process of creating a deb file. It can even upload a package to launchpad for inclusion in the ubuntu reposiotries if you so choose.
+
+**Note:** The package is dependant on the system architecture that you build it on (amd64, amd32 etc). So, ensure that you build it in ubuntu 14.04 64bit or change the architecture value in the control file. You may use a vagrant VM to build this, just use the following command to launch a vm: `vagrant init ubuntu/trusty64 && vagrant up && vagrant ssh`. Also, ensure that you are running the same version of node as that which the packaging installs.
```
devcritter: `amd32` -> `i386`

devcritter:
@rcrichton nice!

Built the package and installation ran perfectly on a clean VM. From my side I definitely think this is a much better approach. Plus the install is really quick now.

Just pointed out a minor type (it's not called amd32), but other than that this is great! :+1:

rcrichton:
Oops, you are right. I'll make that change.

On Fri, 11 Sep 2015 09:35 Hannes Venter <notifications@github.com> wrote:

> @rcrichton <https://github.com/rcrichton> nice!
>
> Built the package and installation ran perfectly on a clean VM. From my
> side I definitely think this is a much better approach. Plus the install is
> really quick now.
>
> Just pointed out a minor type (it's not called amd32), but other than that
> this is great! [image: :+1:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jembi/openhim-core-js/pull/529#issuecomment-139474137>
> .
>
-- 
Ryan Crichton
Lead Developer, Jembi Health Systems |  SOUTH AFRICA
Mobile: +27845829934 | Skype: ryan.graham.crichton
E-mail: ryan@jembi.org

PR #531, opened 2015-08-31
Switch to 1.8.0 release of the pem module

devcritter:
@rcrichton there's a release out of the pem module, and not a moment too soon! the latest master of that module was breaking for us.


rcrichton:
Ok cool. This looks good but should we use `^1.8.0` instead? That seems to be the standard that we have been using. It should be ok because we use shrinkwrap.

devcritter:
Cool, made the switch

PR #544, opened 2015-09-21
Issue #535 OHM-194: Added additional fields to the client model

### LINKED ISSUE #535: Include optional client information in the data model
It would be useful to add additional information about clients to the data model such as:
* organization
* software name
* technical contact person (incl. name and number or email)
* description

devcritter:
Closes #535 

@rcrichton does this look okay to you? Can you think of anything else that would be useful to add?


rcrichton:
@devcritter this looks good to me. The only other thing that I can think of is 'Location'. Do you think that is worth adding? Otherwise, :+1:

devcritter:
Yeah, sounds good to me. Added it in.

rcrichton:
:+1:

PR #546, opened 2015-09-23
Fixes #545: defaultChannelConfig.name sparse index

### LINKED ISSUE #545: Duplicate key error on mediator registration
Occasionally getting this response when starting up a mediator:
```
Response: 500 (Could not add Mediator via the API: MongoError: insertDocument :: caused by :: 11000 E11000 duplicate key error index: openhim-development.mediators.$defaultChannelConfig.name_1  dup key: { : null })
```

devcritter:
@rcrichton this seems to do the trick. the issue is that the channel schema has a unique index on it, and when mediators haven't got default channel config, a null key was added to this index.


rcrichton:
Looks good to me :+1:

PR #551, opened 2015-10-01
Issue540

### LINKED ISSUE #508: Alert on orchestration status
Currently we only trigger alerts on the routes: https://github.com/jembi/openhim-core-js/blob/121d186e1dfb6ab5366b2fba35dce94c3601bc31/src/alerts.coffee#L63-L65

It would be useful to trigger alerts based on orchestrations as well.

### LINKED ISSUE #540: Optimise the alerting query
The mongo query that runs to generate alerts can get very slow and can cause mongo to lock up, it needs to be optimised. Here is an example of the query:

```
query: { $or: [ { response.status: { $gte: 400, $lt: 500 } }, { routes: { $elemMatch: { response.status: { $gte: 400, $lt: 500 } } } } ], channelID: ObjectId('550c2c5febaab96a71b3d4e0'), request.timestamp: { $gte: new Date(1442343696042) } } planSummary: IXSCAN { channelID: 1 }, IXSCAN { request.timestamp: 1 }
```

devcritter:
Closes #540 
Closes #508 

@rcrichton if you could please review?
@armageddon @bausmeier let me know if this looks okay to you guys?

Basically I just made the visualizer TTL collection [more generic](https://github.com/jembi/openhim-core-js/blob/d6f45c9d5dfa71fc2b21276b0bfe4fffa41c754a/src/model/events.coffee) and based the alerts off of it.

I tested locally by running over ten thousand cda validation transactions at 100 per second, and everything coped fine (alerting mongo queries <50ms [under extreme load that is, in general it runs 1-2 ms])


src/alerts.coffee
```diff
@@ -60,19 +61,21 @@ getAllChannels = (callback) -> Channel.find {}, callback
 findGroup = (groupID, callback) -> ContactGroup.findOne _id: groupID, callback
 
 findTransactions = (channelID, dateFrom, status, callback) ->
```
rcrichton: Perhaps rename this to `findEvents()` or `findTxEvents()`

rcrichton: Actually never mind, I see it projects transactionIds.

src/middleware/events.coffee
```diff
@@ -90,66 +95,58 @@ storeVisualizerEvents = (ctx, done) ->
...
+    event: 'end'
+    name: ctx.authorisedChannel.name
+    status: ctx.response.status
+    visualizerStatus: status
 
   now = new Date
   event.created = now for event in trxEvents
-  events.VisualizerEvent.collection.ensureIndex { created: 1 }, { expireAfterSeconds: 600 }, ->
-    events.VisualizerEvent.collection.insert trxEvents, (err) -> return if err then done err else done()
+  events.Event.collection.ensureIndex { created: 1 }, { expireAfterSeconds: 3600 }, ->
```
rcrichton: Perhaps we should move this to the schema declaration. I think we can declare it like this: http://mongoosejs.com/docs/api.html#schematype_SchemaType-index

src/middleware/events.coffee
```diff
@@ -90,66 +95,58 @@ storeVisualizerEvents = (ctx, done) ->
...
+    events.Event.collection.insert trxEvents, (err) -> return if err then done err else done()
```
rcrichton: I wonder if we should move back to using mongoose here.

src/model/alerts.coffee
```diff
@@ -6,11 +6,9 @@ Schema = mongoose.Schema
...
   "method":       type: String, required: true
-  "timestamp":    type: Date, required: true, default: Date.now
+  "timestamp":    type: Date, required: true, default: Date.now, expires: '1d'
   "channelID":    type: String, required: true
   "status":       type: String, required: true
-  "transactions": [String]
-  "error":        String
   "alertStatus":  type: String, required: true, enum: ["Failed", "Completed"]
  
 exports.Alert = connectionDefault.model 'Alert', AlertSchema
```
rcrichton: Is this collection only used to determine if alerts have previously been sent to a user? Perhaps, add a brief comment about what the collection is used for. I had forgotten.

rcrichton:
Hey @devcritter this looks great to me. Good idea! I have a few minor comment otherwise :+1:

src/middleware/events.coffee
devcritter: Afraid this is necessary if we're doing inserts like below since we're basically bypassing the mongoose schema

src/middleware/events.coffee
devcritter: It's apparently much better in terms of performance to do it this way[*](http://stackoverflow.com/questions/16726330/mongoose-mongodb-batch-insert)[*](https://groups.google.com/forum/#!msg/mongoose-orm/IkPmvcd0kds/bZuYCN_orckJ). It looks like the only method that mongoose has for doing batch insert is `.create`, but it sounds like that basically just loops and calls `.save` for each document, rather than doing a proper batch insert.

src/model/alerts.coffee
devcritter: Aah, yeah it is. Sorry yeah definitely need to comment it! I forgot too when I first saw this now :)

src/middleware/events.coffee
rcrichton: Ok, fair enough.

rcrichton:
@devcritter one super minor comment, otherwise its good to merge :+1:

PR #563, opened 2015-10-09
Migrated versioning page from openhim.org

devcritter:
@rcrichton just migrated the versioning page from openhim.og. Can update badge links to readthedocs.


rcrichton:
Nice! But, maybe this should be in the user-guide not the dev-guide? It affects users installing the software. What do you think?

devcritter:
Aah, yeah you're right, that would be better

PR #564, opened 2015-10-09
Update versioning.md

devcritter:
oops - had the links wrong way round


rcrichton:
Ah sorry I didn't even notice when I looked at it.

PR #572, opened 2015-10-26
Add actor start and stop audits

### LINKED ISSUE #571: Audit log OpenHIM application start

rcrichton:
Closes #571

### DIFF
--- a/src/api/mediators.coffee
+++ b/src/api/mediators.coffee
@@ -4,8 +4,10 @@ Q = require 'q'
 logger = require 'winston'
 authorisation = require './authorisation'
 semver = require 'semver'
+atna = require 'atna-audit'
 
 utils = require "../utils"
+auditing = require '../auditing'
 
 
 exports.getAllMediators = ->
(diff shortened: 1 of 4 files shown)


rcrichton:
@devcritter could you have a look at this one as well. The problem here is that the startup and shutdown audit will get logged once per cluster worker. I can't find a better way to do this and technically that may be the correct thing to do. What do you think?

devcritter:
@rcrichton cool, looks good to me - I just haven't had a chance to look into the audit library yet; but I'll make any comments regarding in the other PR so that it's in one place.

I'm not sure what the best practice is wrt ATNA, but it seems alright to me to have every worker audit - rather more than less I think. Besides if someone clustered over mutliple servers - it would anyway have audited multiple times - which seems correct to me.

Otherwise all good :+1: although some of the issues in the other PR might also apply here...

devcritter:
Oh I almost forgot: On the ticket there's a discussion about also auditing mediator startups when they register. Do you want to take this as part of this PR?

rcrichton:
Ok cool. I'l take a look at adding mediator registration auditing. That shouldn't be too hard.

rcrichton:
@devcritter Ok, I've made the changes and added start auditing for mediators as well.

src/api/mediators.coffee
```diff
@@ -53,6 +55,18 @@ exports.addMediator = ->
...
   try
     mediator = this.request.body
+
+    if mediator?.endpoints?[0]?.host?
+      mediatorHost = mediator.endpoints[0].host
+    else
+      mediatorHost = 'unknown'
+
+    # audit mediator start
+    audit = atna.appActivityAudit true, mediator.name, mediatorHost, 'root'
```
devcritter: to be fair, we don't actually know that it's the root user. Not sure if there's a way to pick up the user dynamically - couldn't see anything from a quick search - maybe rather call it `system` instead? just to make it a clearer

devcritter:
Looks good. Just a minor comment about using the root user if you agree, but otherwise :+1:

src/api/mediators.coffee
rcrichton: Yeah, I just put root there for lack of something better. `system` does sound better. I'l change it.

PR #573, opened 2015-10-26
User audit

### LINKED ISSUE #569: Audit log the fact that an OpenHIM user as logged in
The OpenHIM should audit the fact that a user has logged in (or maybe even viewed certain transactions) with it internal audit log.

rcrichton:
Closes #569

### DIFF
--- a/src/api/users.coffee
+++ b/src/api/users.coffee
@@ -34,7 +34,7 @@ exports.authenticate = (email) ->
         ts: new Date()
 
       # User authenticated, send audit
-      audit = atna.userLoginAudit atna.OUTCONE_SUCCESS, 'openhim', os.hostname(), email, user.groups.join(','), user.groups.join(',')
+      audit = atna.userLoginAudit atna.OUTCOME_SUCCESS, 'openhim', os.hostname(), email, user.groups.join(','), user.groups.join(',')
       audit = atna.wrapInSyslog audit
       auditing.processAudit audit, -> logger.info 'Processed internal audit'
   catch e
--- a/src/api/users.coffee
+++ b/src/api/users.coffee
@@ -9,7 +9,9 @@ contact = require '../contact'
 config = require "../config/config"
 config.newUserExpiry = config.get('newUserExpiry')
 utils = require "../utils"
-
+atna = require 'atna-audit'
+os = require 'os'
+auditing = require '../auditing'
 
 ###
 # Get authentication details
@@ -22,10 +24,19 @@ exports.authenticate = (email) ->
 
     if not user
       utils.logAndSetResponse this, 404, "Could not find user by email #{email}", 'info'
+      # User NOTE authenticated, send audit
+      audit = atna.userLoginAudit 8, 'openhim', os.hostname(), email
+      audit = atna.wrapInSyslog audit
+      auditing.processAudit audit, -> logger.info 'Processed internal audit'
     else
       this.body =
         salt: user.passwordSalt
         ts: new Date()
+
+      # User authenticated, send audit
+      audit = atna.userLoginAudit 0, 'openhim', os.hostname(), email, user.groups.join(','), user.groups.join(',')
+      audit = atna.wrapInSyslog audit
+      auditing.processAudit audit, -> logger.info 'Processed internal audit'
   catch e
     utils.logAndSetResponse this, 500, "Error during authentication #{e}", 'error'
 
@@ -62,7 +73,7 @@ exports.getNewUser = (token) ->
 exports.updateNewUser = (token) ->
   token = unescape token
   userData = this.request.body
-  
+
   try
     # first try get new user details to check expiry date
     newUserOldData = yield User.findOne(token: token).exec()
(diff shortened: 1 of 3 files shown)


rcrichton:
@devcritter could you please review this for me. You may want to have a look at the ATNA module code as well. Its also new: https://github.com/jembi/atna-audit

I just need to work out why a module won't install on travis which is causing the test to fail.

rcrichton:
I had to add g++ for libxmljs to compile, also I have dropped support for node 0.12. What do you think of this? I needed to drop this to use modules which use ES6 features. (my ATNA module)

src/api/users.coffee
```diff
@@ -22,10 +24,19 @@ exports.authenticate = (email) ->
 
     if not user
       utils.logAndSetResponse this, 404, "Could not find user by email #{email}", 'info'
+      # User NOT authenticated, send audit
+      audit = atna.userLoginAudit 8, 'openhim', os.hostname(), email
```
devcritter: magic numbers: it'd be good to define the outcome codes somewhere - perhaps even in your atna audit lib. (handy [ref](https://github.com/jembi/uct-redcap/blob/master/acl-portal/backend/ATNA.coffee#L8) to just copy/paste from)

src/api/users.coffee
```diff
@@ -22,10 +24,19 @@ exports.authenticate = (email) ->
...
+      audit = atna.wrapInSyslog audit
+      auditing.processAudit audit, -> logger.info 'Processed internal audit'
     else
       this.body =
         salt: user.passwordSalt
         ts: new Date()
+
+      # User authenticated, send audit
+      audit = atna.userLoginAudit 0, 'openhim', os.hostname(), email, user.groups.join(','), user.groups.join(',')
```
devcritter: magic numbers

devcritter:
@rcrichton do you mean core would now always need a C++ compiler in order to install? Is there no other way around this? (we once [dropped a lib](https://github.com/jembi/openhim-core-js/issues/387) because of this same issue...).

No issue from my side in dropping support for node 0.12 - but from our hipchat discussions, I think we need consensus wrt the major version number.

devcritter:
Haven't had a chance to look at the atna audit lib - will make relevant comments here as soon as I get a chance

devcritter:
Another note: in the ticket we mention `"or maybe even viewed certain transactions"` - from my side I definitely think this would be great to implement, perhaps we could have a flag on a channel `auditViews`? Or would rather take this on as part of another ticket?

src/api/users.coffee
rcrichton: Cool, good idea.

rcrichton:
Yeah it would mean that we need a c++ compiler. I'm not super happy about this either. I can't find a pure js way to do schema validation. I could remove this from the atna module and just leave the tests in that module to do validation. That means that the user won't get runtime validation but maybe that isn't essential. It's probably better to not require a c++ compiler for windows environments as mentioned in that issue.

I think the view transactions should be taken on in another issue. It would take a bit more thought.

rcrichton:
@devcritter, ok I've made the changes. I removed the libxmljs module from the atna module dependencies. Its now just a dev-dependency. I also removed all ES6 features from the ATNA module so that we can continue supporting Node v0.12 a bit more. Seeming as it's a general purpose lib I think this is best for now instead of virally requiring projects to move to Node v4.

devcritter:
Looks good :+1: Cool, yeah it's definitely much better to keep the dependency on a lower version.

Also had a look at the audit util lib - looks good, just one comment: I see you're using RFC rather than the DICOM format. Any reason not to use DICOM? Switching shouldn't take more than renaming some of the attributes...

src/api/users.coffee
```diff
@@ -22,10 +24,19 @@ exports.authenticate = (email) ->
...
+      audit = atna.userLoginAudit atna.OUTCOME_SERIOUS_FAILURE, 'openhim', os.hostname(), email
+      audit = atna.wrapInSyslog audit
+      auditing.processAudit audit, -> logger.info 'Processed internal audit'
     else
       this.body =
         salt: user.passwordSalt
         ts: new Date()
+
+      # User authenticated, send audit
+      audit = atna.userLoginAudit atna.OUTCONE_SUCCESS, 'openhim', os.hostname(), email, user.groups.join(','), user.groups.join(',')
```
devcritter: typo: outcone

bausmeier: @rcrichton "OUTCONE" looks like a typo.

rcrichton:
Cool, I fixed the typo.

I found it difficult to find the latest DICOM format spec for the ATNA module. I thought that RFC was the latest. I see there is this, which extends the RFC. I'l work on adding some of the changes in the ATNA module. In the mean time is this PR good to go?

devcritter:
Yup - good to go :+1: 

Yeah I believe DICOM is the latest specified by IHE - it's pretty simple to switch though - it's just some of the attribute names ([e.g.](https://github.com/jembi/openhim-core-js/blob/master/src/auditing.coffee#L9-L10)). The technical framework had a nice section detailing the differences ([page 154](http://www.ihe.net/uploadedFiles/Documents/ITI/IHE_ITI_TF_Vol2a.pdf)).

PR #575, opened 2015-10-28
Logs api

### LINKED ISSUE #558: Providing an API endpoint to get the server logs would be useful
Related to https://github.com/jembi/openhim-console/issues/334

rcrichton:
Closes #558 

@devcritter could you review this for me. I decided to add a winston transport to log to mongo which the API uses, this could be generally useful for the OpenHIM for closing inspection of issues. What do you think about this?

### DIFF
--- a/src/api/logs.coffee
+++ b/src/api/logs.coffee
@@ -1,0 +1,19 @@
+logger = require 'winston'
+moment = require 'moment'
+Q = require 'q'
+
+exports.getLogs = ->
+  query = this.request.query
+  if not query?
+    query = {}
+
+  options =
+    from: query.from || moment().subtract(5, 'minutes').toDate()
+    until: query.until || new Date
+    order: 'asc'
+    start: query.start || 0
+    limit: query.limit || 1000000 # limit: 0 doesn't work :/
+
+  results = yield Q.ninvoke logger, 'query', options
+  this.body = results.mongodb
+  this.status = 200
--- a/src/api/logs.coffee
+++ b/src/api/logs.coffee
@@ -4,6 +4,12 @@ Q = require 'q'
 authorisation = require './authorisation'
 utils = require "../utils"
 
+levels =
+  debug: 1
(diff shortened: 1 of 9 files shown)


src/api/logs.coffee
```diff
@@ -0,0 +1,26 @@
+logger = require 'winston'
+moment = require 'moment'
+Q = require 'q'
+authorisation = require './authorisation'
+utils = require "../utils"
+
+exports.getLogs = ->
```
devcritter: It might be good to think a bit about how this endpoint will be used from the console with regard to filters. Two uses I can think of would be to filter by date range as well as filter by log level.

In addition it would be cool if the console page continually updates with the latest log entries - I think for all we would need is to supply the from date, leaving the until as the latest time. (the from would be the console's last sync date - this is how the transaction log updates work)

devcritter:
@rcrichton looks good to me! :+1: Tested locally and it worked well.

devcritter:
Hmmm thinking about this a bit more. I don't know if I'm so too crazy about storing logs in the database - the db size is going to grow really quickly, esp. with those scheduled tasks logging every minute. I see winston-mongodb has an option to use a capped collection. Should we enable something like that?

rcrichton:
Yeah it currently does have filters, the only isssue is that log level is something that is set on the winston transport. I could go with debug level on the mongodb transport and then try filter by the desired level on the query, but then we would be storing even more logs in the db.

I thought about using a capped collection, but thought it may be useful for some instances to have access to all the historic logs. Perhaps this should be enabled by default thought.

devcritter:
Cool, yeah I would suggest that it needs to be as configurable as possible - perhaps add config options for using capped and perhaps even an option for disabling the transport for implementers that would rather not log to the db?

rcrichton:
@devcritter I've made the changes, let me know what you think.

devcritter:
:+1: Looks good

PR #585, opened 2015-11-10
Fixes #545: Refactored channels schema definition

### LINKED ISSUE #545: Duplicate key error on mediator registration

devcritter:
Closes #545 for real this time

@rcrichton if you could please review? The problem was that the index definition in the channel schema was propagated through to the mediator schema when it reused it - i refactored it a bit so as to keep the index definition out of the channel def. The mediator collection therefore won't even have an index on channel.name anymore, which i think is correct.

### DIFF
--- a/src/model/channels.coffee
+++ b/src/model/channels.coffee
@@ -30,8 +30,8 @@ RewriteRuleDef =
   "toPort":         type: Number, required: true, default: 80
   "pathTransform":  String
 
-ChannelSchema = new Schema
-  "name":               type: String, required: true, unique: true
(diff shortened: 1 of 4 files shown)


rcrichton:
Ah ok, that makes sense. :+1:

PR #586, opened 2015-11-10
Auditing: changed logging to debug level

devcritter:
@rcrichton a quick one. This logging gets a bit hectic with lots of mediators all sending heartbeats!


rcrichton:
Cool, that's fine by me :+1:

PR #590, opened 2015-11-12
Mediators: Added support for 'map' config type

devcritter:
@rcrichton what do you think?

There's a lot we can do e.g. with regards to the dynamic config for the file queue but I think this map is already really useful and would be great to keep as is. Therefore I think whatever we look into next should be done in addition to this as another config type/plugin architecture/whatever.

### DIFF
--- a/docs/dev-guide/mediators.md
+++ b/docs/dev-guide/mediators.md
@@ -78,6 +78,7 @@ The `configDefs` property defines an array of configuration definitions that eac
 * `bool` - A boolean value (true or false)
 * `number` - An integer or decimal value
 * `option` - A value from a pre-defined list. If this datatype is use then the `values` property MUST also be used. The `values` property specifies an array of possible values for the parameter.
+* `map` - Key/value pairs. A map is formatted as an object with string values, e.g. `{ "key1": "value1", "key2": "value2" }`. New keys can be added dynamically.
 
 The OpenHIM-core SHALL respond with a HTTP status of 201 if the mediator registration was successful.
 The OpenHIM-core SHALL respond with an appropriate 4xx status if the mediator registration could not be completed due to a bad request.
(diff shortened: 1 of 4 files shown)


test/unit/mediatorsAPITest.coffee
```diff
@@ -111,3 +111,76 @@ describe "Mediator API unit tests", ->
...
+            k1: "v1"
+            k2: "v2"
+          ]
+        )
+      catch err
+        return
+
+      throw new Error 'Failed'
+
+    it "should reject config that includes a 'map' that isn't an object", ->
```
rcrichton: Perhaps, change the test names slightly so we know which is failing as this one and the one above are exactly the same. You could add '(array)' and '(string)' to the end of the name.

test/unit/mediatorsAPITest.coffee
```diff
@@ -111,3 +111,76 @@ describe "Mediator API unit tests", ->
...
+          param1:
+            k1: "v1"
+            k2: 42
+        )
+      catch err
+        return
+
+      throw new Error 'Failed'
+
+    it "should reject config that includes a 'map' with non-string values", ->
```
rcrichton: Duplicate name - same as above.

docs/dev-guide/mediators.md
```diff
@@ -78,6 +78,7 @@ The `configDefs` property defines an array of configuration definitions that eac
 * `bool` - A boolean value (true or false)
 * `number` - An integer or decimal value
 * `option` - A value from a pre-defined list. If this datatype is use then the `values` property MUST also be used. The `values` property specifies an array of possible values for the parameter.
+* `map` - Key/value pairs. A map is formatted as an object with string values, e.g. `{ "key1": "value1", "key2": "value2" }`. New keys can be added dynamically.
```
rcrichton: How are new keys added? Do you have to supply a template object in the initial default config? Are these added via the UI?

rcrichton: I see now, that it is both. Maybe change the Last sentence to "New keys/value pairs can be added dynamically".

rcrichton:
Looks good :+1:, just a few minor comments.

test/unit/mediatorsAPITest.coffee
devcritter: good point

PR #600, opened 2015-11-25
Config struct repeats

### LINKED ISSUE #591: Add support for grouping a set of mediator config fields together
It would be useful to group fields together under a particular category. Eg. 'DHIS detail' or 'Details for file queue # 1'

### LINKED ISSUE #592: Add support for repeating mediator config fields
This would be useful for repeating elements like list of names. It would al,so be useful to allow group of fields to be repeated (see https://github.com/jembi/openhim-core-js/issues/591).

devcritter:
Closes #591 
Closes #592 

For repeating fields, rather that something like `repeats: 3` I instead opted to define it as `array: true`. I think it's a simpler option, but for now there is no max limit on the size of the array. I think if it comes up, we can figure out a solution for it, but I think it's easier to work with like this.


rcrichton:
@devcritter very nice, adding all the example to the documentation is great! The code all looks good, I just need to test this with console.

rcrichton:
:+1: this works great with the console PR.

PR #606, opened 2016-01-19
Fix the cert used in testing

rcrichton:
The certs expired which caused the tests to start to fail with weird
errors. +50year of validaity should do the trick!


devcritter:
:+1:

PR #607, opened 2016-01-19
Audit node auth failure

### LINKED ISSUE #609: Audit node authentication failures

rcrichton:
@devcritter please review.

The one issue that I have here is that authentication could be passed through to a different mechanisms, ie basic auth and still pass authentication. I have no way of telling that though. So, I simply report and error whenever the TLS middleware can't authenticate, even if the client will be authenticated down the line. I think this is ok, as it's still a node authentication failure. However, it's not ideal. Thoughts?

Closes #609

### DIFF
--- a/src/middleware/authorisation.coffee
+++ b/src/middleware/authorisation.coffee
@@ -3,9 +3,11 @@ Q = require "q"
 xpath = require "xpath"
 dom = require("xmldom").DOMParser
 logger = require "winston"
+atna = require 'atna-audit'
 config = require '../config/config'
 config.authentication = config.get('authentication')
 utils = require '../utils'
+auditing = require '../auditing'
 
 statsdServer = config.get 'statsd'
 application = config.get 'application'
@@ -64,6 +66,11 @@ extractContentType = (ctHeader) ->
     return ctHeader.substring(0, index).trim()
   else
     return ctHeader.trim()
+    
+genAuthAudit = (remoteAddress) ->
+  audit = atna.nodeAuthentication remoteAddress, 'openhim', os.hostname(), atna.OUTCOME_MINOR_FAILURE
+  audit = atna.wrapInSyslog audit
+  return audit
 
 # export private functions for unit testing
 # note: you cant spy on these method because of this :(
@@ -73,6 +80,7 @@ if process.env.NODE_ENV == "test"
   exports.matchXpath = matchXpath
   exports.matchJsonPath = matchJsonPath
   exports.extractContentType = extractContentType
+  exports.genAuthAudit = genAuthAudit
 
 # Is the channel enabled?
 # If there is no status field then the channel IS enabled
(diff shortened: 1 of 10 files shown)


rcrichton:
Nice, looks good @devcritter :+1:

devcritter:
Would it be possible to move the TLS middleware to after the basic auth? That way if it fails it will definitely have failed completely.

rcrichton:
Hey @devcritter, it would be possible I think. But the middleware is really designed to be generic so we can add more authorisation middleware. So it wouldn't really help in that design, although, we are probably not likely to add more. Also, I'd rather keep it trying the strongest form of auth first.

What if I add a check if the client supplied a cert. If they did and auth failed, log the failure. If they didn't don't audit. Although this may miss cases where a cert is not supplied nor is basic auth.

Another option is to just add a config option for this logging...

devcritter:
Sure yeah, fully agree with keeping the design generic.

How about adding a "deferred" error handler? Where the audting is only attempted after all the mechanisms have been tried. Something like:
```
app.use tlsAuth
app.use basicAuth
app.use authAuditor
```

then in tlsAuth it has something like
```
if failed then ctx.authErr = 'oh no!'
```
instead of immediately auditing.

authAuditor can then handle the audit after all mechanisms have completed
```
if not ctx.authenticated
  audit ctx.authErr
else
  audit 'success'
```

rcrichton:
That sounds like a good idea. I'm not sure if the audit will really make sense for basic auth (it's of type 'node authentication'), but it's probably better than nothing. I'l give that a go, thanks.

rcrichton:
@devcritter I've moved this to the authorisation middleware as then I can support all types of authentication failure and also not audit if the endpoint is public. Let me know what you think.

src/middleware/authorisation.coffee
```diff
@@ -64,6 +66,11 @@ extractContentType = (ctHeader) ->
     return ctHeader.substring(0, index).trim()
   else
     return ctHeader.trim()
+    
+genAuthAudit = (remoteAddress) ->
+  audit = atna.nodeAuthentication remoteAddress, 'openhim', os.hostname(), atna.OUTCOME_MINOR_FAILURE
```
devcritter: See #610 - you can lookup the audit source id from config now

src/middleware/authorisation.coffee
```diff
@@ -128,6 +136,7 @@ exports.authorise = (ctx, done) ->
     if config.authentication.enableBasicAuthentication
       ctx.set "WWW-Authenticate", "Basic"
     logger.info "The request, '" + ctx.request.path + "', is not authorised to access any channels."
+    auditing.sendAuditEvent genAuthAudit(ctx.ip), -> logger.info 'Processed nodeAuthentication audit'
```
devcritter: I've been using logger.debug for the audits, mostly because it fills up the logs with a lot of noise (e.g. auths from mediator heartbeats).

I know this is probably a low volume audit and it's probably not the same issue, but maybe for consistency we should keep them on debug?

devcritter:
@rcrichton nice, looks good! I like the idea of auditing on all auth failures like you're doing now - seems like good practice.

Just some minor comments, but otherwise good to go :+1:

rcrichton:
@devcritter could you please have another quick look.

devcritter:
@rcrichton looks good :+1:

PR #610, opened 2016-01-20
ATNA: Added option for setting the audit source id

devcritter:
@rcrichton a quick one: I moved the audit source id into config


rcrichton:
Nice, looks good :+1:

PR #617, opened 2016-02-08
OHM-229 #612: Added channel priority functionality

### LINKED ISSUE #612: Channel Orders
Currently when multiple channels match an incoming transaction, the first found channel is used. To better manage this, a `order` field should be added to channels. With this then if multiple channels match, then the one with lower order should be used.

hnnesv:
Closes #612 

### DIFF
--- a/src/middleware/authorisation.coffee
+++ b/src/middleware/authorisation.coffee
@@ -74,20 +74,10 @@ genAuthAudit = (remoteAddress) ->
   audit = atna.wrapInSyslog audit
   return audit
 
-# export private functions for unit testing
-# note: you cant spy on these method because of this :(
-if process.env.NODE_ENV == "test"
-  exports.matchContent = matchContent
-  exports.matchRegex = matchRegex
-  exports.matchXpath = matchXpath
-  exports.matchJsonPath = matchJsonPath
-  exports.extractContentType = extractContentType
-  exports.genAuthAudit = genAuthAudit
 
 # Is the channel enabled?
 # If there is no status field then the channel IS enabled
 exports.isChannelEnabled = isChannelEnabled = (channel) -> not channel.status or channel.status is 'enabled'
-
 
 channelWithHighestPriority = (channels) ->
   channels.reduce (channel1, channel2) ->
@@ -165,3 +155,15 @@ exports.koaMiddleware = (next) ->
   if this.authorisedChannel?
     sdc.timing "#{domain}.authorisationMiddleware", startTime if statsdServer.enabled
     yield next
+
+
+# export private functions for unit testing
+# note: you cant spy on these method because of this :(
+if process.env.NODE_ENV == "test"
+  exports.matchContent = matchContent
+  exports.matchRegex = matchRegex
+  exports.matchXpath = matchXpath
+  exports.matchJsonPath = matchJsonPath
+  exports.extractContentType = extractContentType
+  exports.genAuthAudit = genAuthAudit
+  exports.channelWithHighestPriority = channelWithHighestPriority
--- a/src/middleware/authorisation.coffee
+++ b/src/middleware/authorisation.coffee
@@ -6,6 +6,7 @@ logger = require "winston"
 atna = require 'atna-audit'
 config = require '../config/config'
 config.authentication = config.get('authentication')
+config.logger = config.get('logger')
 utils = require '../utils'
 auditing = require '../auditing'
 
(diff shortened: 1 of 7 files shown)


src/middleware/authorisation.coffee
```diff
@@ -88,56 +88,72 @@ if process.env.NODE_ENV == "test"
...
+              else
+                # deny access to channel if the content type isnt set
+                return #(continue loop)
+
+            # now check that the status is 'enabled' and if the message content matches
+            if isChannelEnabled(channel) and matchContent(channel, ctx.body)
+              approvedChannels.push channel
+
+    if approvedChannels.length > 0
+      # authorisation succeeded
```
rcrichton: Perhaps add a debug statement that lists all matched channels. For debugging priority.

hnnesv:
@rcrichton added it in

test/integration/e2eIntegrationTests.coffee
```diff
@@ -1027,4 +1027,106 @@ describe "e2e Integration Tests", ->
...
+        request("http://localhost:5001")
+          .get("/test/undefined/priority")
+          .auth("testApp", "password")
+          .expect(200)
+          .end (err, res) ->
+            if err
+              done err
+            else
+              res.text.should.be.exactly 'target1' #should route to target1 via channel2
+              done()
```
rcrichton: I think that we need a test here for channels with identical priority.

src/middleware/authorisation.coffee
```diff
@@ -88,56 +88,72 @@ if process.env.NODE_ENV == "test"
 exports.isChannelEnabled = isChannelEnabled = (channel) -> not channel.status or channel.status is 'enabled'
 
 
+channelWithHighestPriority = (channels) ->
```
rcrichton: This needs to be unit tested thoroughly so that we can verify all it's behaviours.

rcrichton:
@hnnesv this looks good. I'm a bit worried about how we ensure that channels always have a different priority but I guess I have to see how this works in console. Is it anticipated that channels always have a different priority or are overlaps expected?

I've added a few more comments above as well.

test/integration/e2eIntegrationTests.coffee
hnnesv: Yeah I thought about adding that - but i wonder if the behaviour shouldn't be considered random if the priorities are equal (or both none)? Of course the logical assumption would be that the first added channel should be picked (and that's how the check was written), but I'm beginning to think that there shouldn't be any guarantee from our side about the behaviour (why should insert-order matter?).

What do you think? Or do you think we should have clearly defined behaviour in this case?

hnnesv:
@rcrichton oh does the debug logs get captured even if we're not on debug level? that's fine then, I was just trying to avoid the string-concat and map when it's not needed.

hnnesv:
@rcrichton well I think it's really a configuration issue for implementors. It all depends on your channels: if you've got two channels `/patient` and `/encounter` - you probably won't care about priotities since there's not conflict. But if you have `/patient`, `/encounter` and `/.*` then the priory will allow you to resolve conflicts. You could either make 1, 2 and 3. Or make `/patient` 1, `/encounter` 1 and `/.*` 2 - the priority is conflict resolver.

rcrichton:
@hnnesv yeah it does for the server logs in mongo. This allows the user to switch between info and debug dynamically and view past logs. Everything from debug up is captured in mongo.

rcrichton:
@hnnesv yeah I'm just wondering how this would work in practice in console, we wouldn't want users to deal with priority numbers directly. It should be drag and drop. In that case I think updating each channel to have a different number would be best.

hnnesv:
@rcrichton @BMartinos's changes are ready if want to give it a go on your system? https://github.com/jembi/openhim-console/pull/363

I thought the approach is working quite nicely. Having to rely on a global view of the channels where we need to ensure that they are in order makes it more complicated I think, especially at an API level.

hnnesv:
(fyi travis is driving me nuts! it seems to be the same error everytime, but the only change since the last build on the PR is the debug statement. and the tests works locally)

hnnesv:
@rcrichton added in the requested unit tests if you could take another look. What do you think about the general strategy now after seeing the console?

test/integration/e2eIntegrationTests.coffee
rcrichton: I simply though that it seem there is order implied by how the code is written so we should test for that.

rcrichton:
@hnnesv cool, this looks good to me. I like how it's turned out now in console and here. It's simple yet effective. :+1:

PR #618, opened 2016-02-09
OHM-225 Issue #613: Routes enable/disable

### LINKED ISSUE #613: Disable/enable routes
It should be possible to enable/disable specific routes

hnnesv:
Closes #613 

### DIFF
--- a/src/middleware/router.coffee
+++ b/src/middleware/router.coffee
@@ -22,11 +22,17 @@ os = require 'os'
 domain = "#{os.hostname()}.#{application.name}.appMetrics"
 sdc = new SDC statsdServer
 
-containsMultiplePrimaries = (routes) ->
+
+isRouteEnabled = (route) -> not route.status? or route.status is 'enabled'
+
+exports.numberOfPrimaryRoutes = numberOfPrimaryRoutes = (routes) ->
   numPrimaries = 0
   for route in routes
-    numPrimaries++ if route.primary
-  return numPrimaries > 1
+    numPrimaries++ if isRouteEnabled(route) and route.primary
+  return numPrimaries
+
+containsMultiplePrimaries = (routes) -> numberOfPrimaryRoutes(routes) > 1
+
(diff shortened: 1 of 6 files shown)


src/middleware/router.coffee
```diff
@@ -22,11 +22,17 @@ os = require 'os'
 domain = "#{os.hostname()}.#{application.name}.appMetrics"
 sdc = new SDC statsdServer
 
-containsMultiplePrimaries = (routes) ->
+
+isRouteEnabled = (route) -> not route.status? or route.status is 'enabled'
+
+exports.numberOfPrimaryRoutes = numberOfPrimaryRoutes = (routes) ->
```
rcrichton: I wonder if this would be better implemented as a mongoose virtual: http://mongoosejs.com/docs/guide.html#virtuals

It seems more like a virtual property of a channel.

rcrichton:
@hnnesv looks good. Just a refactoring comment above.

src/middleware/router.coffee
hnnesv: @rcrichton I took a look, but ran into issues with the unit tests. Some of them were failing after I added it in, and I traced the problem to the fact that the virtual binding got "lost" at some point. I think it's a problem with how the schema module (i.e. the actual model/channels file) gets loaded when running the tests - I had the same issue with the indexes (e.g. [here](https://github.com/jembi/openhim-core-js/blob/master/test/unit/alertsTest.coffee#L147) we need to reload the indexes even though they're declared at schema level). I'm sure we can get it to work, but it seems a bit clunky (at least for our codebase) to implement atm.

rcrichton:
@hnnesv Ok sure, that doesn't sound great for our codebase. Could you write an issue describing this so we can try address it. It shouldn't hold us up when there is a cleaner why of implementing things.

Anyway, with regards to the PR, this isn't a blocking issue so :+1:

PR #625, opened 2016-02-11
OHM-226 #599: Only add report data if channel is enabled or has data

### LINKED ISSUE #599: Disabled channels should not appear in reports

hnnesv:
Closes #599 

### DIFF
--- a/src/reports.coffee
+++ b/src/reports.coffee
@@ -84,8 +84,8 @@ sendReports = (job, flag, done) ->
               # If report has been fetched get it from the map
               if channelReportMap[channel._id]
                 data = channelReportMap[channel._id]
-                # add report - always add if the channel is enabled, otherwise only if there is data
-                if data.channel.status is 'enabled' or data.data.length isnt 0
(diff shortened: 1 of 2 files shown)


src/reports.coffee
```diff
@@ -83,7 +83,10 @@ sendReports = (job, flag, done) ->
 
               # If report has been fetched get it from the map
               if channelReportMap[channel._id]
-                reportMap[userKey].data.push channelReportMap[channel._id]
+                data = channelReportMap[channel._id]
+                # add report - always add if the channel is enabled, otherwise only if there is data
+                if data.channel.status is 'enabled' or data.data.length isnt 0
```
rcrichton: Is channel status always set? Just thinking of channels that were have never had their status explicitly set (they existed before the functionality was added to core). Can status ever be empty?

rcrichton:
Hey @hnnesv, look good. Just a question above. Also is their any way to add tests to this?

hnnesv:
@rcrichton added a check. There are no existing tests and would need refactoring since the method sends out emails

rcrichton:
Ok sure :+1:

PR #645, opened 2016-03-01
Issue #644: Copy location header

### LINKED ISSUE #644: Location header not copied
When the response status isn't 3xx, then the location header still doesn't get copied from upstream to client. This is needed for FHIR where a [location header needs to be returned](https://www.hl7.org/fhir/http.html#create) on a create.

hnnesv:
Closes #644 

@rcrichton if you could please review? the location header wasn't copied correctly


rcrichton:
Cool, looks good to me :+1:

PR #656, opened 2016-03-03
Add common root CAs to our https server

### LINKED ISSUE #655: We should trust common root CAs when authenticating clients
Right now we poison the CA list of our https server with only the certificates that we trust from our keystore. This mean that TLS verification will fail for any certificate that we don't have the full certificate chain (up to the the root CA) for. It seems it is common for root CAs not to give out their root certificates and rather rely on intermediate certificates to prevent them from having to re-issue all certificates if they happen to be compromised.

Thus, it is impossible to authenticate with the OpenHIM with a certificate signed by one of the CAs where the root CA isn't available. This means that for the most part only self signed chains can be used.

To solve this we could add the common root CAs back to the OpenHIM's https server using something like https://github.com/coolaj86/node-ssl-root-cas#rootcas

rcrichton:
Closes #655

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "request": "^2.51.0",
     "semver": "^5.1.0",
     "source-map-support": "^0.4.0",
+    "ssl-root-cas": "^1.1.10",
     "statsd-client": "^0.2.1",
     "winston": "^2.2.0",
     "winston-mongodb": "^1.2.0",
--- a/src/middleware/tlsAuthentication.coffee
+++ b/src/middleware/tlsAuthentication.coffee
@@ -5,6 +5,7 @@ Keystore = require("../model/keystore").Keystore
 logger = require "winston"
 utils = require '../utils'
 pem = require 'pem'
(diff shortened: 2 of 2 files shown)


rcrichton:
@hnnesv please could you review this for me?

hnnesv:
@rcrichton the problem / solution makes sense to me :+1: 

[**minor usability note**] the module's console.log's are quite ugly when you first startup the him and then it tells you to restart the server, which I can imagine being a bit annoying when you're deploying the HIM on a new server / or for new users. Perhaps it would be possible to trigger the download with npm install by having a simple js script that just does a `require('ssl-root-cas/latest').rootCas`? triggering the download at install time rather than when you first startup the HIM? Really not a big deal though - can leave as future work if you haven't got space or if it's tricky

rcrichton:
Yeah it is a bit ugly. I'm not sure it's really necessary to do more work on right now. Even without the restart it will use a the set of CAs that the package automatically comes with. I think this is ok for now, not sure we get much more form putting effort into this.

hnnesv:
Cool :+1:

PR #677, opened 2016-03-23
Mhero roles

### LINKED ISSUE #647: Client roles should be managed separately from users
Allow Roles to be created and stored separately from clients. This will allow the user to physically create roles and assign roles to users rather than just entering something when creating a user.

hnnesv:
Closes #647 

Will keep PR open for end-to-end testing with https://github.com/jembi/openhim-console/pull/374 and https://github.com/jembi/openhim-console/pull/373 to see if it meets our needs


docs/dev-guide/api-ref.md
```diff
@@ -153,9 +153,163 @@ where `:clientId` is the `_id` property of the client to delete.
...
+    }
+  ]
+}
+```
+
+#### Add a new role
+
+`POST /roles`
+
+with a json body containing the role name and channels and clients to apply to. Clients do not need to be specified, but at least one channel has to. Channels and clients can be specified either by their `_id` or `name` for a channel and `clientID` for a client.
```
rcrichton: If there are clients present does a channel still need to be added? If so, then I'm not sure if this is the best rule. Often users would want to create user and roles before assigning them to channels. At least that is how I would do it.

src/api/roles.coffee
```diff
@@ -0,0 +1,248 @@
...
+# Roles is a virtual API; virtual in the sense that it is not linked
+# to a concrete roles collection.
+#
+# Rather it an abstraction of the 'allow' field on Channels and 'roles' on Clients,
+# providing a mechanism for setting up allowed permissions.
+###
+
+
+filterRolesFromChannels = (channels, allClients) ->
+  roles = {}
```
rcrichton: I'm guessing an object is used initially to make sure we don't have duplicates. Perhaps a comment to this effect would be useful to the reader.

docs/dev-guide/api-ref.md
rcrichton: From reading the code is it not that either a client or a channel need to be added? I may be missing something.

rcrichton:
Hey @hnnesv this is looking good. I just have some comment about the inner working of this.

src/api/roles.coffee
```diff
@@ -15,7 +15,7 @@ utils = require '../utils'
 
 
 filterRolesFromChannels = (channels, allClients) ->
-  roles = {}
+  rolesMap = {} # K: permission, V: channels that share permission
```
rcrichton: I think this will only return roles that have channels associated with them. Wouldn't we need to search clients as well. Especially since this is used to get all roles in line 49.

src/api/roles.coffee
```diff
@@ -0,0 +1,251 @@
...
+
+exports.getRole = (name) ->
+  # Test if the user is authorised
+  if not authorisation.inGroup 'admin', this.authenticated
+    return utils.logAndSetResponse this, 403, "User #{this.authenticated.email} is not an admin, API access to getRole denied.", 'info'
+
+  try
+    result = yield Channel.find({allow: {$in: [name]}}, {name: 1 }).exec()
+    if result is null or result.length is 0
+      utils.logAndSetResponse this, 404, "Role with name '#{name}' could not be found.", 'info'
```
rcrichton: Won't this miss channels that just have clients?

src/api/roles.coffee
```diff
@@ -0,0 +1,251 @@
...
+    this.status = 500
+
+
+exports.deleteRole = (name) ->
+  # Test if the user is authorised
+  if not authorisation.inGroup 'admin', this.authenticated
+    return utils.logAndSetResponse this, 403, "User #{this.authenticated.email} is not an admin, API access to updateRole denied.", 'info'
+
+  try
+    result = yield Channel.find({allow: {$in: [name]}}, {name: 1 }).exec()
```
rcrichton: I think you need to check clients here as well.

test/integration/rolesAPITests.coffee
```diff
@@ -431,6 +431,31 @@ describe 'API Integration Tests', ->
...
+          .expect(201)
+          .end (err, res) ->
+            return done err if err
+            Client.find roles: $in: ['role4'], (err, clients) ->
+              return done err if err
+              clients.length.should.be.exactly 2
+              mapId = (arr) -> (arr.map (a) -> "#{a._id}")
+              mapId(clients).should.containEql "#{client1._id}"
+              mapId(clients).should.containEql "#{client2._id}"
+              done()
```
rcrichton: I think we need test to check if we can query/update roles with just clients, no channel. Maybe this is just a change in the test data.

rcrichton:
@hnnesv a few comments above. I find this a little difficult to read so hopefully I'm understanding correctly.

src/api/roles.coffee
hnnesv: oops yeah it would

test/integration/rolesAPITests.coffee
hnnesv: Already done for update, see `should add a role for clients` below. Will add for query.

hnnesv:
@rcrichton ready for review

rcrichton:
Thanks @hnnesv, looks good. Just the tests are failing. If you can get them passing I've move to done.

hnnesv:
Cool fixed it, we're back to our usual travis issues :) it's an out of scope change though

rcrichton:
Hmmm, ok. Any idea why the change was needed?

rcrichton:
:+1:

hnnesv:
No idea :/

PR #691, opened 2016-04-13
Changed to relative links for _static resources

hnnesv:
Hey @rcrichton looks like the resource links are broken

Not sure if this is the correct way to specify them! This works when I build the docs locally though.

Thanks to @psmatsinhe for pointing these out


rcrichton:
@hnnesv I think this looks fine, it seems that images work the way this was previously. I'm not sure why these don't. Lets try this fix and see what happens on read the docs.

hnnesv:
Okay it works now. Yeah not sure why it was working for images and not for this.

PR #701, opened 2016-04-22
Add windows how-to

hnnesv:
@rcrichton this is a win install guide.

@cseebregts could you please take a read through this as well? thanks


rcrichton:
Thanks @hnnesv, looks good to me. :+1:

cseebregts:
Thanks so much @hnnesv, I followed your instructions and was successfully able to install the OpenHIM Core and Console. I used IIS instead of Nginx. Will submit a PR with IIS instructions as I think most windows users will use IIS as a web server.

PR #702, opened 2016-04-25
MHER-40 authorisation refactor

rcrichton:
@hnnesv please could you review this for me. It's a fairly big change, but I think it improves readability and maintainability greatly.


src/middleware/authorisation.coffee
```diff
@@ -19,134 +15,57 @@ os = require 'os'
...
   audit = atna.nodeAuthentication remoteAddress, himSourceID, os.hostname(), atna.OUTCOME_MINOR_FAILURE
   audit = atna.wrapInSyslog audit
   return audit
 
+authoriseClient = (channel, ctx) ->
+  if ctx.authenticated? and channel.allow?
+    if ctx.authenticated.roles?
+      match = false
+      channel.allow.forEach (role) ->
+        if ((ctx.authenticated.roles.indexOf role) isnt -1)
```
hnnesv: I know this was existing code, but this might be a nice opportunity to refactor this in a more coffee'ish way:
```
if role not in ctx.authenticated.roles
```

src/middleware/authorisation.coffee
```diff
@@ -19,134 +15,57 @@ os = require 'os'
...
+          match = true
+          return
+      if match
+        return true
+    if ((channel.allow.indexOf ctx.authenticated.clientID) isnt -1)
```
hnnesv: ditto above refactoring comment

src/middleware/authorisation.coffee
```diff
@@ -19,134 +15,57 @@ os = require 'os'
...
+      return true
+
+  return false
+
+authoriseIP = (channel, ctx) ->
+  if channel.whitelist?.length > 0
+    return (channel.whitelist.indexOf ctx.ip) isnt -1
```
hnnesv: ditto above refactoring comment

src/middleware/authorisation.coffee
```diff
@@ -19,134 +15,57 @@ os = require 'os'
...
-      logger.info "The request, '" + ctx.request.path + "' is authorised to access " + ctx.authorisedChannel.name
-    else
-      # authorisation failed
-      ctx.response.status = 401
-      if config.authentication.enableBasicAuthentication
-        ctx.set "WWW-Authenticate", "Basic"
-      logger.info "The request, '" + ctx.request.path + "', is not authorised to access any channels."
-      auditing.sendAuditEvent genAuthAudit(ctx.ip), -> logger.debug 'Processed nodeAuthentication audit'
+  channel = ctx.matchingChannel
+  if channel? and (channel.authType == 'public' || isAuthorised(channel, ctx))
```
hnnesv: rather use `or` instead of `||`

src/middleware/authorisation.coffee
```diff
@@ -19,134 +15,57 @@ os = require 'os'
...
+    # authorisation succeeded
+    ctx.authorisedChannel = channel
+    logger.info "The request, '" + ctx.request.path + "' is authorised to access " + ctx.authorisedChannel.name
```
hnnesv: refactoring opportunity: `"The request, #{ctx.request.path} ..."`

src/middleware/authorisation.coffee
```diff
@@ -19,134 +15,57 @@ os = require 'os'
...
+  else
+    # authorisation failed
+    ctx.response.status = 401
+    if config.authentication.enableBasicAuthentication
+      ctx.set "WWW-Authenticate", "Basic"
+    logger.info "The request, '" + ctx.request.path + "', is not authorised to access any channels."
```
hnnesv: refactoring opportunity: as above

src/middleware/requestMatching.coffee
```diff
@@ -0,0 +1,137 @@
...
+  if channel.matchContentRegex
+    return matchRegex channel.matchContentRegex, ctx.body
+  else if channel.matchContentXpath and channel.matchContentValue
+    return matchXpath channel.matchContentXpath, channel.matchContentValue, ctx.body
+  else if channel.matchContentJson and channel.matchContentValue
+    return matchJsonPath channel.matchContentJson, channel.matchContentValue, ctx.body
+  else if channel.matchContentXpath or channel.matchContentJson
+    # if only the match expression is given, deny access
+    # this is an invalid channel
+    logger.error 'Channel with name "' + channel.name + '" is invalid as it has a content match expression but no value to match'
```
hnnesv: refactoring opportunity: `"Channel with name #{..}"`

src/middleware/requestMatching.coffee
```diff
@@ -0,0 +1,137 @@
...
+
+matchUrlPattern = (channel, ctx) ->
+  pat = new RegExp channel.urlPattern
+  return pat.test ctx.request.path
+
+matchContentTypes = (channel, ctx) ->
+  if channel.matchContentTypes?.length > 0
+    if ctx.request.header and ctx.request.header['content-type']
+      ct = extractContentType ctx.request.header['content-type']
+      if (channel.matchContentTypes.indexOf ct) >= 0
```
hnnesv: refactoring opportunity: `if ct in channel.matchContentTypes`

src/middleware/authorisation.coffee
```diff
@@ -19,134 +15,57 @@ os = require 'os'
...
-    if not channel1.priority? then return channel2
-    if channel1.priority <= channel2.priority then return channel1
-    return channel2
+authFunctions = [
+  authoriseClient,
+  authoriseIP
+]
 
+isAuthorised = (channel, ctx) ->
+  return authFunctions.some (authFunc) ->
```
hnnesv: hmmm i don't think this is correct - so in this case _anyone_ who has a whitelisted IP is authenticated?

The whitelisting, if enabled, should happen _in addition_ to the authoriseClient check

src/middleware/requestMatching.coffee
```diff
@@ -0,0 +1,137 @@
...
+matchChannel = (channel, ctx) ->
+  matchFunctions.every (matchFunc) ->
+    return matchFunc channel, ctx
+
+findMatchingChannel = (channels, ctx) ->
+  return channels.find (channel) ->
+    return matchChannel channel, ctx
+
+matchRequest = (ctx, done) ->
+  utils.getAllChannels (err, channels) ->
```
hnnesv: Maybe for readability just add comment stating that utils.getAllChannels will have sorted the channels according to priority - or better yet maybe rename the method to something like `getAllChannelsPrioritySorted`

test/integration/e2eIntegrationTests.coffee
```diff
@@ -1288,3 +1289,29 @@ describe "e2e Integration Tests", ->
             else
               res.text.should.be.exactly 'target1' #should route to target1 via channel2
               done()
+
+    it "should deny access if multiple channels match but the top priority channel denys access", (done) ->
```
hnnesv: typo: denys -> denies

hnnesv:
@rcrichton definitely way more readable now!

Some comments above.

src/middleware/authorisation.coffee
hnnesv: (late thought) the public check should first check the whitelist

hnnesv:
(added a late comment regarding the public check)

src/middleware/authorisation.coffee
rcrichton: I'm not sure if it should be. A public endpoint is accessible to anyone isn't it? The whitelist works with private channels that are authenticated by IP only.

hnnesv: That's not how it's currently implemented and that was my understanding of it too - whitelist means a client must have a approved IP in all cases, even for public.

Thinking about it I also think this approach also makes sense as it offers more implementation options:
Fully public (no whitelist+public), public for approved hosts (whitelist+public), authenticated for world (no whitelist+private), authenticated for approved hosts (whitelist+private).

If we change now then we're cutting out the option to have authenticated approved hosts.

rcrichton: Ok, that makes sense. So, you are advocating for when there is a whitelist set then ONLY those hosts can access that channel and IN ADDITION any other authorisation checks must also apply. Is that right?

That seems to be how it is implemented now, I must have misread it...

hnnesv: Yup that's exactly it.

rcrichton:
@hnnesv, thanks. Please review again.

hnnesv:
Looks good!

PR #707, opened 2016-05-06
Add IIS Instructions

cseebregts:
Hi @hnnesv, I have added IIS instructions and consolidated the how-to. Please, could you check to see if you happy with the consolidation?

### DIFF
--- a/docs/how-to/how-to-install-on-windows.md
+++ b/docs/how-to/how-to-install-on-windows.md
@@ -40,9 +40,17 @@ or with whichever file location you chose to create for the config.
 
 ## OpenHIM Console
 
-### Install Nginx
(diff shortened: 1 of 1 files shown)


docs/how-to/how-to-install-on-windows.md
```diff
@@ -40,9 +40,17 @@ or with whichever file location you chose to create for the config.
 
 ## OpenHIM Console
 
-### Install Nginx
+A web server will be required to host the OpenHIM Console and in this guide we will use IIS and Nginx. However any good web server will be suitable, e.g. Apache.
```
hnnesv: Hmmm it might be a bit confusing with both sets of instructions now (_"do i need both iis aaand nginx?"_). Maybe we should put the nginx instructions at the bottom with note saying it's an alternative?

hnnesv:
Thanks @cseebregts!

Just a note above

hnnesv:
Thanks again @cseebregts

PR #713, opened 2016-05-11
OHM-243 Apply a factor to setTimeout calls in testing

rcrichton:
The factor is increased when travis is detected to be running the tests.
This allows us to have quick setTimeout values when testing locally,
however, these are dilated when run on travis. Travis is slower due to
it being a virtual environment.


hnnesv:
👍 Looks good to me

PR #729, opened 2016-06-08
OHM-246: Audits - added index for event timestamp

hnnesv:
OHM-246

That easy :) tested on uct staging and saw an immediate improvement


rcrichton:
Whew that was a lot to review! ;) :+1:

PR #733, opened 2016-06-09
OHM-248: Move auditing into authentication middleware

hnnesv:
Add auditing exemption paths

OHM-248

### DIFF
--- a/src/api/authentication.coffee
+++ b/src/api/authentication.coffee
@@ -3,6 +3,23 @@ crypto = require 'crypto'
 logger = require 'winston'
 config = require "../config/config"
 config.api = config.get('api')
+config.auditing = config.get('auditing')
+atna = require 'atna-audit'
+auditing = require '../auditing'
+os = require 'os'
+himSourceID = config.auditing.auditEvents.auditSourceID
+
+# will NOT audit any successful logins on the following paths (specified as regex patterns)
+# only 'noisy' endpoints should be included, such as heartbeats or endpoints that get polled
+auditingExemptPaths = [
+  /\/transactions/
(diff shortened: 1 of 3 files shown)


src/api/authentication.coffee
```diff
@@ -3,6 +3,23 @@ crypto = require 'crypto'
...
+config.auditing = config.get('auditing')
+atna = require 'atna-audit'
+auditing = require '../auditing'
+os = require 'os'
+himSourceID = config.auditing.auditEvents.auditSourceID
+
+# will NOT audit any successful logins on the following paths (specified as regex patterns)
+# only 'noisy' endpoints should be included, such as heartbeats or endpoints that get polled
+auditingExemptPaths = [
+  /\/transactions/
```
rcrichton: Should transaction be included here? They may be the most important to audit.

rcrichton:
@hnnesv looks good. Just a question above.

src/api/authentication.coffee
hnnesv: It is one the endpoints that gets polled though by console. What do you think - should we allow it then regardless? It is quite noisy

rcrichton: Yeah, the only problem is that we are possibly disclosing PHI when we query for transactions. Does the polling use a simple view of the transactions as a request param that we could match on?

hnnesv: Yup it does - yeah good idea. We can add a check to only exempt if on a simple view.

hnnesv:
@rcrichton addressed the /transactions case

rcrichton:
Nice looks good :+1:

PR #745, opened 2016-06-29
Adds a client ID header to forwarded requests

rcrichton:
@hnnesv could you please review this? I can't figure out an easy way to test this without it being more trouble than the actual code change. Let me know what you think.


hnnesv:
👍 Cool yeah, no doesn't make sense to spend a lot of effort testing such a small change. Looks good!

PR #746, opened 2016-07-07
Retry requests

hnnesv:
OHM-250

PR is opened for design purposes: https://docs.google.com/a/jembi.org/document/d/1cG95FGB8lQsGm5ZEP_yOb4D1vWPhazQJw1cW3LXv_m4/edit?usp=sharing

It's working but is probably not yet ready for merging


src/autoRetry.coffee
```diff
@@ -0,0 +1,83 @@
...
+    for channel in results
+      do (channel) ->
+        deferred = Q.defer()
+
+        findTransactions channel, (err, results) ->
+          if err
+            logger.error err
+            deferred.resolve()
+          else if results? and results.length>0
+            createRerunTask (results.map((r) -> r._id)), -> deferred.resolve()
```
rcrichton: I wonder if we should rather create 1 task with transactions from all channels, that way we could not overload the server with multiple tasks for each channel?

src/model/transactions.coffee
```diff
@@ -38,17 +38,19 @@ RouteMetadataDef =
...
+  "childIDs":             [Schema.Types.ObjectId]
+  "channelID":            type: Schema.Types.ObjectId, index: true
+  "request":              RequestDef
+  "response":             ResponseDef
+  "routes":               [RouteMetadataDef]
+  "orchestrations":       [OrchestrationMetadataDef]
+  "properties":           Object
+  "canRerun":             type: Boolean, default: true
+  "wasRerun":             type: Boolean, default: false
+  "internalServerError":  Boolean
```
rcrichton: I think for the real implementation we should handle internal errors better. Both in the core and console. I think storing the error text would be very useful as well as handling of mediator returning internal errors.

src/model/channels.coffee
```diff
@@ -33,34 +33,36 @@ RewriteRuleDef =
...
+  "txViewAcl":              [String]
+  "txViewFullAcl":          [String]
+  "txRerunAcl":             [String]
+  "alerts":                 [AlertsDef]
+  "status":                 type: String, default: 'enabled', enum: ['enabled', 'disabled', 'deleted']
+  "rewriteUrls":            type: Boolean, default: false
+  "addAutoRewriteRules":     type: Boolean, default: true
+  "rewriteUrlsConfig":      [RewriteRuleDef]
+  "autoRetryEnabled":       type: Boolean, default: false
+  "autoRetryPeriodMinutes": type: Number, default: 60, min: 1
```
rcrichton: I think we need to implement a retry max number of times for this to be used in production.

rcrichton:
@hnnesv, generally this is looking good. For the real implementation I'd like to see the above mentioned changes. I'm going to break these down into sub-tasks on jira.

src/model/transactions.coffee
```diff
@@ -59,3 +61,14 @@ TransactionSchema.index "request.timestamp"
 
 #compile schema into Model
 exports.Transaction = connectionDefault.model 'Transaction', TransactionSchema
+
+exports.Transaction.collection.createIndex({
+  'request.timestamp': 1
+  channelID: 1
+},
+{
+  name: 'transaction.autoRetry'
```
rcrichton: Why does this need to be a partial index? Should we not just permanently add these indexes for all transactions? Both seems useful?

src/autoRetry.coffee
hnnesv: Atm it excludes transactions in the search that are too new (`to` date in search) based on the autoRetryPeriod, which is a channel specific setting. I'm sure it's possible to use a single task though; but it will make the logic a bit more complex.

src/model/channels.coffee
hnnesv: It had that in the original design, but left it out for simplicity. It would also be good to discuss the workflow a bit I think; e.g. sending out alerts if transactions have reached max tries etc.

src/model/transactions.coffee
hnnesv: Yeah mediator handling will be essential

src/autoRetry.coffee
rcrichton: Ok cool. I don't think it would be too hard using closures to keep  track of the `tx._id`s then creating a single task.

rcrichton: Just a note for the final implementation of this.

src/model/transactions.coffee
hnnesv: The partial index is built on the expression, so it will only include transactions where internalServerError is true (and wasRerun false). It basically a handle on all those transactions that need to be rerun; so I don't think it would be useful in a generic way.

PR #750, opened 2016-07-13
Visualizer mediators

hnnesv:
OHM-127

Companion PR https://github.com/jembi/openhim-console/pull/394

I refactored the events api a little bit and deprecated the visualizer specific api.

I also noticed that events aren't being generated for secondary routes (if those routes complete after the primary). This must've broken when we did the route async upgrade. I wanted to try pick that up here; but have run out of time - so I'll make another ticket for that.


config/config.md
```diff
@@ -117,19 +117,10 @@ The following config option are provided by the OpenHIM. All of these options ha
...
-    // Should orchestration and route timestamps be normalized to the
-    // transaction time. This may be useful if you have mediators running on
-    // other servers
-    "enableTSNormalization": true,
-    // A buffer period for orchestration timestamps
-    "orchestrationTSBufferMillis": 100
+  "events": {
+    // Should event timestamps be normalized to the transaction time.
+    // This will be useful if you have mediators running on other servers
+    "enableTSNormalization": true
```
rcrichton: I'm not quite sure I understand what this means.

src/api/visualizer.coffee
```diff
@@ -1,36 +1,33 @@
 Event = require('../model/events').Event
```
rcrichton: Maybe rename this file to events?

rcrichton: Oh, if it's deprecated maybe we should just remove it entirely. The next release is going to be a breaking one anyways.

src/koaApi.coffee
```diff
@@ -93,7 +94,9 @@ exports.setupApp = (done) ->
   app.use route.put '/tasks/:taskId', tasks.updateTask
   app.use route.delete '/tasks/:taskId', tasks.removeTask
 
+  # DEPRECATED use /events/:receivedTime
   app.use route.get '/visualizer/events/:receivedTime', visualizer.getLatestEvents
+  # DEPRECATED use /heartbeat
   app.use route.get '/visualizer/sync', visualizer.sync
```
rcrichton: Again, maybe we should just remove entirely.

src/middleware/events.coffee
```diff
@@ -3,7 +3,12 @@ logger = require 'winston'
 events = require '../model/events'
 messageStore = require '../middleware/messageStore'
 config = require "../config/config"
-config.visualizer = config.get('visualizer')
+config.events = config.get('events')
+
+if !config.events
+  # maybe we're using outdated config
+  config.events = config.get('visualizer')
```
rcrichton: Should this not be `config.get('events')`

src/middleware/events.coffee
```diff
@@ -3,7 +3,12 @@ logger = require 'winston'
...
+  config.events.normalizationBuffer = config.events.orchestrationTsBufferMillis
```
rcrichton: Does 'orchestrationTsBufferMillis' exist anymore in config?

rcrichton:
Hey @hnnesv this looks good to me. I've added a few comments. Also, I think we should just completely remove the visualiser API instead of deprecating. We are going to be doing a breaking change release for next release anyways due to the metrics API, may as well include this.

In testing this I am seeing mediator fail events not showing on the mediator itself but rather on core itself. I haven't tracked that down yet, maybe you have some ideas?

config/config.md
hnnesv: If a mediator is running on another server, the reported timestamps don't always line up with core - so what this does is normalize those timestamps relative to core - ie. if a mediator orchestration takes 200ms, the events will be set relative to core's start time (orchestration event start == core start and orchestration event end == core start + 200ms), rather than the "real" orchestration time.

In an ideal world all servers would be running on the same time, but unfort that's just not always the case, even with ntp enabled, they might drift slightly before syncing up again. I think this was the case when we added this in (i believe a couple mins before a demo!!).

I think even a few ms out made for ugly results on the visualizer.

src/middleware/events.coffee
hnnesv: Oh, what I was trying to do here was backwardly supporting the older config format: `if !newConfig then tryReadOldFormat`. I guess if we're removing the old visualizer API, then there's no reason to do this though. What do you think? Keeping it will make upgrading easier for people though?

src/middleware/events.coffee
hnnesv: See above comment re: backwards compatibility

hnnesv:
@rcrichton cool, that's fine - will remove that API. Will look into the mediator error...

hnnesv:
Afraid I can't reproduce the error issue, this is what I'm getting:
screen shot 2016-07-22 at 12 42 05

hnnesv:
Ready for review

rcrichton:
Hmm ok, I'l check again my side. Maybe my mediator isn't returning a good response.

hnnesv:
One thing you can check if you haven't already is if the mediator is returning `x-mediator-urn` - that's how we identify mediator events

rcrichton:
Seems my mediator wasn't responding quite right. So the problem was on my end...

src/middleware/events.coffee
rcrichton: If it makes upgrading easier then I'm ok to keep it.

rcrichton:
LGTM :+1:

PR #756, opened 2016-07-25
Ohm 256 async events

hnnesv:
OHM-256


rcrichton:
Thanks @hnnesv LGTM :+1:

PR #757, opened 2016-07-25
Metrics: timestamp fix

hnnesv:
@rcrichton a quick bug fix - seems moment (annoyingly) starts months on 0, so the timestamps were one month ahead in the results.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "koa-request": "^1.0.0",
     "koa-route": "^2.4.2",
     "koa-statsd": "^0.0.2",
+    "lodash": "^4.14.0",
     "moment": "^2.12.0",
     "mongoose": "^4.4.18",
     "nconf": "^0.8.4",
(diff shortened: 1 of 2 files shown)


rcrichton:
Ah oops, I did think that I must double check this at some stage and then subsequently forgot...

LGTM :+1: Just may want to check the tests.

hnnesv:
@rcrichton hmmm looks like the tests are failing under node 0.12. I'm happy to investigate, but maybe this is a good opportunity to raise the question of whether we still want to support 0.12? I'm personally in favour of dropping - what do you think?

bausmeier:
0.12 is technically still a supported version of Node https://github.com/nodejs/LTS if that counts for anything.

hnnesv:
Aah right. Thanks @bausmeier - yeah I think we need to take that under consideration.

hnnesv:
Ok cool, traced the issue - we can avoid the question for today :)

@rcrichton if you could take a glance again please?

rcrichton:
I also think we can drop node 0.12. However, we would need to update the packaging to install node 4 rather. Right now I think it's stuck on 0.12.7.

Cool, I'l take a look.

rcrichton:
LGTM :+1:

PR #759, opened 2016-07-29
OHM-266, OHIE-131: Metadata API

napiergit:


docs/dev-guide/api-ref.md
```diff
@@ -784,3 +784,99 @@ returns 200 ok with `{ master: <core-uptime>, mediators: { <urn>: <mediator-upti
...
+    {
+      model: 'Channel'
+      record: { ChannelObject1 }
+      status: 'Valid'
+      message: 'Ok'
+      uid: ChannelObject1.name
+    }, {
+      model: 'Client'
+      record: { ClientObject1 }
+      status: 'Conflict'
```
rcrichton: I'mm guessing this one under successes should  actuall be Valid?

docs/dev-guide/api-ref.md
```diff
@@ -784,3 +784,99 @@ returns 200 ok with `{ master: <core-uptime>, mediators: { <urn>: <mediator-upti
...
+      message: 'Ok'
```
rcrichton: Should there just be no `message` attribute on success?

docs/dev-guide/api-ref.md
```diff
@@ -784,3 +784,99 @@ returns 200 ok with `{ master: <core-uptime>, mediators: { <urn>: <mediator-upti
...
+
+`Post /metadata/validate` Returns `201` and an object with the following format:
+
+```js
+{
+  "successes": [
+    {
+      model: 'Channel'
+      record: { ChannelObject1 }
+      status: 'Valid'
```
rcrichton: Is there a distinction between separating `successes` and `errors` and the `status` property. It seems they relay the same information. I'm wondering if one or the other should be done away with.

docs/dev-guide/api-ref.md
```diff
@@ -784,3 +784,99 @@ returns 200 ok with `{ master: <core-uptime>, mediators: { <urn>: <mediator-upti
...
+    },
+    //...
+  ],
+  "errors": [
+    {
+      model: 'Client'
+      record: { ClientObject2 }
+      status: 'Error'
+      message: e.message
+      uid: ClientObject2.ClientId
```
rcrichton: Maybe include actual example of these rather than the code behind them.

docs/dev-guide/api-ref.md
napiergit: Agreed

src/api/metadata.coffee
```diff
@@ -0,0 +1,161 @@
...
+    return
+
+  try
+    exportObject = {}
+    params = this.request.query
+    
+    # Return all documents from all collections for export
+    for col of collections
+      exportObject[col] = yield collections[col].find().lean().exec()
+      for doc of exportObject[col]
```
rcrichton: I think you are meaning to use `for ... in array` here and line 69 instead of `for ... of`. After that line 72 and 73 could probably be `if doc._id`. Unless I'm missing something.

napiergit: I need the for...of for line 70 exportObject[col]

docs/dev-guide/api-ref.md
napiergit: Yes this is possibly too confusing.  I'll give it some thought for a minute.  could separate them into valid, error, conflict

src/api/metadata.coffee
```diff
@@ -0,0 +1,161 @@
...
+          
+        catch e
+          logger.error "Failed to insert #{key} with unique identifier #{uid}. #{e.message}"
+          returnObject.errors.push buildResponseObject key, doc, 'Error', e.message, uid
+        
+    this.body = returnObject
+    this.status = 201
+  catch e
+    this.body = e.message
+    utils.logAndSetResponse this, 500, "Could not fetch all users via the API #{e}", 'error'
```
rcrichton: The code between this function and the previous one is basically identical. It would be better to combine these into one function so that the code isn't repeated. Remember the DRY principle :)

test/integration/metadataAPITests.coffee
```diff
@@ -0,0 +1,327 @@
...
+          .set("auth-salt", authDetails.authSalt)
+          .set("auth-token", authDetails.authToken)
+          .send(sampleMetadata)
+          .expect(201)
+          .end (err, res) ->
+            if err
+              done err
+            else
+              res.body.errors.length.should.equal 0
+              res.body.successes.length.should.equal 5
```
rcrichton: You should check that the metadata was actually inserted into mongo.

test/integration/metadataAPITests.coffee
```diff
@@ -0,0 +1,327 @@
...
+          .set("auth-salt", authDetails.authSalt)
+          .set("auth-token", authDetails.authToken)
+          .send(testMetadata)
+          .expect(201)
+          .end (err, res) ->
+            if err
+              done err
+            else
+              res.body.errors.length.should.equal 1
+              res.body.successes.length.should.equal 4
```
rcrichton: Again check in mongo.

test/integration/metadataAPITests.coffee
```diff
@@ -0,0 +1,327 @@
...
+            else
+              done()         
+                
+                
+                
+                
+                
+                
+                
+
```
rcrichton: `removeProperties` and `getUniqueIdentifierForCollection` should be unit tested in their own unit test file as well. Those are good examples of testable functions.

docs/dev-guide/api-ref.md
```diff
@@ -784,3 +784,99 @@ returns 200 ok with `{ master: <core-uptime>, mediators: { <urn>: <mediator-upti
 Returns the server uptime in seconds. Includes a list of all registered mediators and if heartbeats have been received for them, will include their uptimes as well.
 
 Note that this is a public endpoint that does not require any authorization. It is convenient for integrating with external monitoring tools.
+
+### Metadata resource
+
+The metadata resource allows the user to export and import Channels, Clients, Mediators, Users and ContactGroups.
+
+#### Export Metadata
```
rcrichton: Perhaps add a little more about when each how each API endpoint should be used. I.e. validate before sending the main import request.

src/api/metadata.coffee
rcrichton: But exportObject[col] is an array right? So you could iterate over the array using `in`. `of` is for object properties or is there something I'm missing.

rcrichton:
Looking good @napiergit, a few comment above. Also please change the name of the PR, it not reallly about fixing the email error anymore, that's more of a side effect.

test/integration/metadataAPITests.coffee
napiergit: This has now been done :)

docs/dev-guide/api-ref.md
```diff
@@ -816,29 +820,25 @@ The metadata resource allows the user to export and import Channels, Clients, Me
 
 ```js
 {
-  "successes": [
+  "rows": [
```
rcrichton: It seems like the `rows` property is unnecessary, we could just return an array instead.

src/api/metadata.coffee
```diff
@@ -19,20 +19,20 @@ collections =
 
 
 #Function to remove properties from export object
-removeProperties = (obj) ->
+exports.removeProperties = (obj) ->
```
rcrichton: Rather than export all of these and make them publicly available to other code, you should rather do something like this - https://github.com/jembi/openhim-core-js/blob/fc9413354df8a9df0377c517dede75d04868d20b/src/middleware/requestMatching.coffee#L131-L137

rcrichton:
@napiergit Looking good! Just a few comments that I see you are already addressing :)

napiergit:
@rcrichton, anything else?

rcrichton:
@napiergit LGTM :+1: 

@nthfloor2 the above change will impact your PR.

PR #766, opened 2016-08-05
OHM-295 get visualizer API endpoint

rcrichton:
This PR includes the changes in #764 as it depends on these. Once that is merged in you will get a truer diff of the changes that are only for the endpoint. For now, you can just review the latest commit.

### DIFF
--- a/src/api/visualizers.coffee
+++ b/src/api/visualizers.coffee
@@ -1,0 +1,15 @@
+Visualizer = require('../model/visualizer').Visualizer
+authorisation = require './authorisation'
+utils = require '../utils'
+
+exports.getAllVisualizers = ->
+  # Must be admin
+  if not authorisation.inGroup 'admin', this.authenticated
+    utils.logAndSetResponse this, 403, 'User #{this.authenticated.email} is not an admin, API access to getAllVisualizers denied.', 'info'
+    return
+
+  try
+    v = yield Visualizer.find().exec()
+    this.body = v
+  catch err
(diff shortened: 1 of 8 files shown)


src/api/visualizers.coffee
```diff
@@ -0,0 +1,15 @@
+Visualizer = require('../model/visualizer').Visualizer
+authorisation = require './authorisation'
+utils = require '../utils'
+
+exports.getAllVisualizers = ->
+  # Must be admin
+  if not authorisation.inGroup 'admin', this.authenticated
+    utils.logAndSetResponse this, 403, 'User #{this.authenticated.email} is not an admin, API access to getAllVisualizers denied.', 'info'
```
hnnesv: String should be using double quotes

src/api/visualizers.coffee
```diff
@@ -0,0 +1,15 @@
...
+    return
+
+  try
+    v = yield Visualizer.find().exec()
+    this.body = v
+  catch err
+    utils.logAndSetResponse this, 500, 'Could not fetch visualizers via the API: #{err}', 'error'
```
hnnesv: ditto: string double quotes

hnnesv:
LGTM. Just some minor comments. There's no need for further review after these and can be merged 👍

src/api/visualizers.coffee
rcrichton: Oops, this was some overzealous fins and replace...

PR #767, opened 2016-08-05
Ohm 279 error results

hnnesv:


napiergit:
@hnnesv, this looks good to me so far.  If you can document the mediator changes it would be very helpful.

src/middleware/messageStore.coffee
```diff
@@ -159,17 +162,18 @@ exports.setFinalStatus = setFinalStatus = (ctx, callback) ->
...
-        callback tx
-
-        if config.statsd.enabled
-          stats.incrementTransactionCount ctx, ->
-          stats.measureTransactionDuration ctx, ->
+      update.status = tx.status
+
+    if ctx.autoRetry? then update.autoRetry = ctx.autoRetry
+
+    if !update then return callback tx # nothing to do
```
napiergit: I don't think this will ever happen because update is defined on line 134?

src/middleware/router.coffee
```diff
@@ -151,6 +161,11 @@ sendRequestToRoutes = (ctx, routes, next) ->
               # handle mediator reponse
               responseObj = JSON.parse response.body
               ctx.mediatorResponse = responseObj
+
+              if responseObj.error?
+                ctx.autoRetry = true
```
napiergit: I don't quite understand why this is here?  I thought we would set on a channel whether or not transactions should be rerun.  This seems to overrule that?

src/middleware/messageStore.coffee
hnnesv: Thanks good catch

src/middleware/router.coffee
hnnesv: So this is saying that the transaction is available to be auto retried - it gives us a handle to quickly query for those transactions that need to be rerun, rather than trying to sift through all failed transactions for a channel

napiergit:
@hnnesv, just minor comment and a question.  But otherwise looks good to me.

src/middleware/router.coffee
hnnesv: It gets picked up here to set flag the transaction: https://github.com/jembi/openhim-core-js/blob/8f892848ef2ed9eb4b851db73c2340b450a6f008/src/middleware/messageStore.coffee#L168

napiergit: Ah ok cool, good explanation.

PR #768, opened 2016-08-05
Ohm 280 single task

hnnesv:
Note that this PR was branch from `ohm-279-error-results` (PR #767)


BMartinos:
Hey @hnnesv the code looks good to me :+1: 
One thing i noticed were some includes which don't seem to be used anywhere within the script.
Line 1,2,10,11

hnnesv:
Thanks, good catch - yeah they can be removed

PR #770, opened 2016-08-08
Ohm 254 add display version

napiergit:


config/about.json
```diff
@@ -0,0 +1,11 @@
...
+  "releases": [
+    {
+      "consoleVersion": "1.7.0",
+      "minimumCoreVersion": "3.0.0"
+    }, {
+      "consoleVersion": "1.6.0",
+      "minimumCoreVersion": "2.0.0"
+    }
+  ]
+}
```
BMartinos: I think that Core shouldn't be in control of which version of console is compatible. This would mean that if changes has been made in console, that it will need to get an entry within the releases config of core as well. 

Ideally i'd think that OpenHIM console should send a request to core only requesting it for its current version. Then in the OpenHIM console config there should be a setting that specifies what the minimum allowed version of Core is that it will work with

src/api/about.coffee
```diff
@@ -0,0 +1,21 @@
...
+aboutInfoObject = require '../../config/about.json'
+
+logger = require 'winston'
+utils = require '../utils'
+authorisation = require './authorisation'
+
+exports.getAboutInformation = () ->
+  
+  if not authorisation.inGroup 'admin', this.authenticated
+    return utils.logAndSetResponse this, 403, "User #{this.authenticated.email} is not an admin, API access to addClient denied.", 'info'
```
BMartinos: I think this check can be removed as all logged in users should be able to see what version is running

BMartinos:
Great thanks @napiergit I think the main comment is just to have the core not reference the console in any way as core can be functional without knowing about the console. Rather let the console only request the version of Core and then Console can set what the minimum version it allows.

BMartinos:
Great thanks @napiergit! LGTM :+1:

PR #772, opened 2016-08-11
Add log rotate script

hnnesv:
@rcrichton a quick one not linked to a ticket. This is just a simple log rotation script. Maybe we should think about bundling this with the deb's?

cc @armageddon


rcrichton:
LGTM :+1: I agree we should add it, do you want to create an issue for it.

PR #775, opened 2016-08-12
OHM-293 visualizer settings migration

rcrichton:


hnnesv:
@rcrichton one issue I've run into is if a user never setup a visualizer, then a visualizer will still be migrated for them, because all user settings are initialized with the default visualizer settings template, i.e. I ended up with
```
{
	"_id" : ObjectId("57b1bd9a77cffe8e034b1c29"),
	"name" : "Super User's visualizer",
	"time" : {
		"maxTimeout" : 5000,
		"maxSpeed" : 5,
		"minDisplayPeriod" : 500,
		"updatePeriod" : 200
	},
	"size" : {
		"padding" : 20,
		"height" : 400,
		"width" : 1000,
		"responsive" : true
	},
	"color" : {
		"text" : "#000000",
		"error" : "#d43f3a",
		"active" : "#4cae4c",
		"inactive" : "#cccccc"
	},
	"mediators" : [ ],
	"channels" : [ ],
	"components" : [ ],
	"__v" : 0
}
```

The visualizer doesn't like having these blank atm and I get errors.

Maybe we should not migrate a visualizer if channels, components and mediators are empty?

src/upgradeDB.coffee
```diff
@@ -62,20 +76,61 @@ upgradeFuncs.push
...
+          promises.push userDefer.promise
+
+          vis = user.settings.visualizer
+          name = "#{user.firstname} #{user.surname}'s visualizer"
+          name = dedupName name, visNames
+          vis.name = name
+          visNames.push name
+
+          vis = new Visualizer vis
+          logger.debug "Migrating visualizer from user profile #{user.email}, using viualizer name '#{name}'"
```
hnnesv: typo: ...using visualizer name...

hnnesv:
@rcrichton it's looking good! Just a couple issues.

rcrichton:
@hnnesv thanks, ya I agree it shouldn't be adding visualizers if those aren't set. I'l make the change.

rcrichton:
@hnnesv please re-review

hnnesv:
👍 Looks good!

PR #787, opened 2016-08-16
Ohm 292 - Visualizer add/edit/clone api endpoints

napiergit:


src/api/visualizers.coffee
```diff
@@ -1,25 +1,102 @@
...
-    v = yield Visualizer.find().exec()
-    this.body = v
+    result = yield Visualizer.findOne(name: name).exec()
+    if not result
+      this.body = "Visualizer with name #{name} could not be found."
+      this.status = 404
+    else
+      this.body = result
   catch err
     utils.logAndSetResponse this, 500, "Could not fetch visualizers via the API: #{err}", 'error'
```
BMartinos: Visualizers should be singular

src/api/visualizers.coffee
```diff
@@ -1,25 +1,102 @@
...
+
+# Endpoint to add new visualizer
+exports.addVisualizer = ->
+  
+  # Must be admin user
+  if not authorisation.inGroup 'admin', this.authenticated
+    return utils.logAndSetResponse this, 403, "User #{this.authenticated.email} is not an admin, API access to addVisualizer denied.", 'info'
+
+  visualizerData = this.request.body
+  if !visualizerData
```
BMartinos: Use 'if not visualizerData'

src/api/visualizers.coffee
```diff
@@ -1,25 +1,102 @@
...
+
+# Endpoint to update specific visualizer by name
+exports.updateVisualizer = (name) ->
+  
+  # Must be admin
+  if not authorisation.inGroup 'admin', this.authenticated
+    return utils.logAndSetResponse this, 403, "User #{this.authenticated.email} is not an admin, API access to updateVisualizer denied.", 'info'
+
+  visualizerData = this.request.body
+  if !visualizerData
```
BMartinos: Use 'if not visualizerData'

src/api/visualizers.coffee
```diff
@@ -1,25 +1,102 @@
...
+    return utils.logAndSetResponse this, 404, "Cannot Update Visualizer with name #{name}, no request object", 'info'
+  
+  name = unescape name
+  
+  # Ignore _id
+  delete visualizerData._id if visualizerData._id
```
BMartinos: Not sure if this is really needed to be deleted?

src/api/visualizers.coffee
```diff
@@ -1,25 +1,102 @@
...
+  
+  try
+    result = yield Visualizer.findOneAndUpdate(name: name, visualizerData).exec()
+    if !result
```
BMartinos: Use 'if not result'

BMartinos:
Great thanks @napiergit! Its looking good! Just a few minor comments above.

Also just want to find out if theres a specific reason for doing the find/update/delete via name and not the mongo ObjectId that gets created?  @rcrichton I've noticed we do this at a few places so not sure if there's a specific reason for this?

Another thing i was thinking about is having the Visualizers visible for non-admin users? They are only allowed to view but not add/edit/delete?

rcrichton:
I hear you guys discussed it and decided to go with using _id. I'm cool with that. The original reason was just to make the endpoint more readable, but _id is probably better.

I agree about non-admin users, but that sounds like a separate issue that we need to think about for the future. Maybe, lets leave that for now.

BMartinos:
Great thanks @rcrichton. Agreed, Lets leave the Visualizer view for non-admin users for another ticket. I think this can also be a bit more complex than just allowing them to pull the list and Visulaizer as they might not have permission to certain channels

BMartinos:
Great thanks @napiergit the changes you made looks good to me :+1:

PR #791, opened 2016-08-22
Packaging: Include build deps in recommends

hnnesv:
@rcrichton I added them as recommends rather depends. Install recommends is true by default anyway and this at least gives the option for implementers who for some reason don't want to enable native compiling.


rcrichton:
Cool good idea. 
LGTM :+1:

PR #796, opened 2016-09-07
OHM-340 autoretry on api update

rcrichton:


napiergit:
Hey @rcrichton, something is not happening correctly when testing this. It's retrying way too many transactions.  Will try to isolate the problem.

napiergit:
The code looks flawless, I think I may have left my channel auto-retry enabled while testing something else, resulting in many failures being rerun.  Trying on a clean db now.

napiergit:
LGTM :+1:

PR #805, opened 2016-10-03
OHM-360: Fix Agenda timezone issue

BMartinos:
Agenda was defaulting to UTC timezone and not using the timezone of the server. Agenda tasks will now executing in the servers' timezone

### DIFF
--- a/src/polling.coffee
+++ b/src/polling.coffee
@@ -28,7 +28,7 @@ exports.registerPollingChannel = (channel, callback) ->
       request options, ->
         done()
 
-    exports.agendaGlobal.every channel.pollingSchedule, "polling-job-#{channel._id}", null, { timezone: utils.serverTimeZone() }
+    exports.agendaGlobal.every channel.pollingSchedule, "polling-job-#{channel._id}", null, { timezone: utils.serverTimezone() }
 
     callback null
 
--- a/src/polling.coffee
+++ b/src/polling.coffee
@@ -5,8 +5,8 @@ config = require './config/config'
 config.polling = config.get('polling')
 logger = require 'winston'
 Q = require 'q'
-logger = require 'winston'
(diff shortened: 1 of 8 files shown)


src/polling.coffee
```diff
@@ -28,7 +28,7 @@ exports.registerPollingChannel = (channel, callback) ->
       request options, ->
         done()
 
-    exports.agendaGlobal.every channel.pollingSchedule, "polling-job-#{channel._id}"
+    exports.agendaGlobal.every channel.pollingSchedule, "polling-job-#{channel._id}", null, { timezone: utils.serverTimezone() }
```
rcrichton: Is there no way to rather set the agenda timezone globally? It seems like it could be error prone to remember to add this whenever we add a new task.

rcrichton: From a quick look it seems like the answer is no...

src/utils.coffee
```diff
@@ -65,3 +66,8 @@ exports.uniqArray = (arr) ->
 
 # thanks to https://coffeescript-cookbook.github.io/chapters/arrays/check-type-is-array
 exports.typeIsArray = Array.isArray || ( value ) -> return {}.toString.call( value ) is '[object Array]'
+
+# get the server timezone
+exports.serverTimezone = () ->
+  timezone = momentTZ.tz.guess()
+  return momentTZ.tz.guess()
```
rcrichton: Either just return `timezone` here or delete line 72.

BMartinos:
Hey @rcrichton I have made the correction as mentioned. Ideally it would be better to have a global place to set the timezone but couldnt find anything in the docs. Seems it needs to be done per agenda task.

napiergit:
@BMartinos LGTM :+1:

PR #812, opened 2016-10-10
OHM-367 Allow certificates to be read from the filesystem on server start

rcrichton:

PR #816, opened 2016-10-11
Ohm 368

BMartinos:
The PR looks up config variables for certificate management and determines if it should enable the file watcher. If the certificate/key is changed, the agenda task is kciked off to restart the server with the new certificate/key.

This PR depends on #812


BMartinos:
Hey @rcrichton ok cool no problem, I think this will make things a lot easier and we wouldnt need to add agenda tasks then. This should be a quick change as well. Will get started on it.

rcrichton:
@BMartinos cool, if you could write a test to ensure that the server is restarted when the cert changes that would be great to.

test/integration/e2eIntegrationTests.coffee
```diff
@@ -1369,3 +1369,34 @@ describe "e2e Integration Tests", ->
...
+            fs.createReadStream('test/resources/certificate-watcher/testBackup.crt').pipe(fs.createWriteStream('test/resources/certificate-watcher/test.crt'));
+
+            Keystore.findOne {}, (err, keystore) ->
+              done(err) if err
+              keystore.cert.data.should.be.exactly 'New certificate content. should trigger certificate reload'
+              done()
+            return
+          ), 1500
+        
+
```
rcrichton: As we chatted about, remove the test until we can figure out a good way to test this... :/

src/server.coffee
```diff
@@ -62,6 +63,22 @@ autoRetry = require './autoRetry'
 
 clusterArg = nconf.get 'cluster'
 
+exports.setupCertificateWatcher = () ->
+  certFile = config.certificateManagement.certPath
+  keyFile = config.certificateManagement.keyPath
+  watcher = chokidar.watch([certFile, keyFile], {
+    usePolling: true
```
rcrichton: Ok cool.

PR #827, opened 2016-10-28
OHM-390: Truncate response bodies if too big

hnnesv:


bausmeier:
Tested and working as expected :+1:

rcrichton:
@hnnesv agreed, nice work.

PR #842, opened 2016-11-11
Fix configuration of router timeout using an environment variable

bausmeier:
The OpenHIM crashed when using an environment variable to configure the router timeout because the value is a string.

To reproduce start the OpenHIM with `router_timeout=120000 npm start` and make a request to the router.

Casting the config value to a number fixes the problem.


bausmeier:
@jembi/openhim Will someone please review this change so that it can go into the next release. I ran into this issue on the project that I'm currently working on so am unable to configure the timeout.

napiergit:
@bausmeier LGTM :+1:

PR #878, opened 2016-12-13
OHM-404: Revert report.type property

hnnesv:


bausmeier:
@hnnesv Please add some tests for this regression. It should be simple to update the sinon assertions in the unit tests to check that the correct field is set.

hnnesv:
@bausmeier actually I started looking into that but noticed that the existing tests where we could put this are quite broken: https://github.com/jembi/openhim-core-js/blob/ec1c7bea22314e329060ec82b643367231f6de5e/test/unit/reportsTest.coffee#L97-L111

(e.g. they're not actually doing anything, done() is called immediately; spies don't seem to be setup properly...) You caught me being too lazy to fix these for a 2-liner :) But I can take a look

bausmeier:
Hmm okay. Maybe we should delete those tests if they aren't actually doing anything. I think that the false sense of confidence is a bad thing. We really need to come up with a strategy for sorting out the testing in the HIM.

bausmeier:
Let's leave it for now though since this story is only half a point. Any idea why the tests are failing as well? I tried rerunning them but it didn't help.

hnnesv:
Ok cool. Yeah fully agree re: false sense of confidence.

Hmmm not sure; the only failures I'm seeing are for the logs API tests which I believe is related to OHM-402. Are these the ones you're seeing?

bausmeier:
I was referring to the tests on Travis. They seem to be failing legitimately. Now that I think about it, it's probably related to the broken pem dependency. I guess that should be fixed separately.

PR #881, opened 2016-12-15
OHM-403 client roles conflict

rcrichton:


bausmeier:
@rcrichton The 0.12 build is failing. I suspect it's because of the use of `Object.assign`. Please will you fix it.

PR #893, opened 2017-02-01
OHM-417: Fix weekly transaction reports

napiergit:


src/reports.coffee
```diff
@@ -23,8 +23,8 @@ sendReports = (job, flag, done) ->
     from = moment().subtract(1, 'days').startOf('day').toDate()
     to = moment().subtract(1, 'days').endOf('day').toDate()
   else
-    from = moment().subtract(1, 'days').startOf('isoWeek').toDate()
-    to = moment().subtract(1, 'days').endOf('isoWeek').toDate()
+    from = moment().startOf('isoWeek').subtract(1, 'weeks').toDate()
+    to = moment().endOf('isoWeek').subtract(1, 'weeks').toDate()
```
napiergit: This was calling the current week

src/reports.coffee
```diff
@@ -111,14 +111,26 @@ sendReports = (job, flag, done) ->
...
+              data.data.forEach (val, index) ->
+                for key, value of totals
+                  if key is 'avgResp'
+                    totals[key] += (if data.data[index]?[key]? then Math.round(data.data[index][key])/1000 else 0)
+                  else
+                    totals[key] += (if data.data[index]?[key]? then data.data[index][key] else 0)
+              
+              for key, value of totals
+                report.data[i][key] = totals[key]
+
```
napiergit: The data.data array returns two items if the week falls over two months. We were only using the first item to build up the weekly report.

src/reports.coffee
rcrichton: @napiergit I don't see how this makes a difference. As it was before it always runs on a Monday due to this - https://github.com/jembi/openhim-core-js/blob/master/config/default.json#L76

I do think your way is a better approach that is more fault tolerant. However, I don't think this would be causing the issue on momconnect unless someone changed the reporting cron schedule to a day other than Monday. Am I missing something?

src/reports.coffee
rcrichton: Awesome :+1:

Would you be able to write a test for this edge case. It seems important to make sure we don't break this again.

napiergit: Yes, good point.  I'll write a unit test.

napiergit:
Actually now that you mention it @Ryan, I am not 100% confident that this solves the issue on momconnect.  It seems to be working locally, but the changes don't seem to relevant to the issue on momconnect.  I'll spend some more time digging.

rcrichton:
@napiergit cool. I do think these changes help but there may be a larger issue still out there.

napiergit:
@rcrichton I am now pretty confident that those changes fix the bug. The aggregation is almost always split over two weeks, which returns two elements in the data.data array.  I will just put in the tests to ensure that this does not get broken again.

rcrichton:
Thanks @napiergit looks good to me!

PR #895, opened 2017-02-23
DAT-84: Enforce max body size on transactions created through api

napiergit:


src/api/transactions.coffee
```diff
@@ -421,7 +448,16 @@ exports.updateTransaction = (transactionId) ->
       if not autoRetryUtils.reachedMaxAttempts tx, channel
         updates.autoRetry = true
         autoRetryUtils.queueForRetry tx
-
+    
+    transactionToUpdate = yield transactions.Transaction.findOne({ _id: transactionId }).exec()
+    bodiesLength = 0
```
BMartinos: Maybe rename this to `defaultBodiesLength` just to indicate that this is the default start for the calculate function

src/api/transactions.coffee
```diff
@@ -201,6 +201,33 @@ exports.getTransactions = ->
...
 
+recursivelySearchObject = (ctx, obj, ws, repeat) ->
+  if Array.isArray obj
+    obj.forEach (value) ->
+      if value && typeof value is 'object'
+        if ws.has(value) then return
+        ws.add(value)
+        repeat(ctx, value, ws)
+  else
+    if obj && typeof obj is 'object'
```
BMartinos: What happens if the type of 'obj' isnt an object, i think we might to add a 'else' statement here and also do a 'return'

test/integration/transactionsAPITests.coffee
```diff
@@ -153,8 +164,8 @@ describe "API Integration Tests", ->
         td = JSON.parse JSON.stringify transactionData
         td.channelID = channel._id
         td.request.body = ''
-        td.response.body = ''
-        td.response.body += '1234567890' for i in [0...2*1024*1024]
+        respBody = largeBody
+        td.response.body = respBody
```
BMartinos: I think we can just use 'largeBody' variable here and remove the variable declaration above

test/integration/transactionsAPITests.coffee
```diff
@@ -177,8 +188,8 @@ describe "API Integration Tests", ->
       it  "should add a transaction and truncate the large request body", (done) ->
         td = JSON.parse JSON.stringify transactionData
         td.channelID = channel._id
-        td.request.body = ''
-        td.request.body += '1234567890' for i in [0...2*1024*1024]
+        reqBody = largeBody
+        td.request.body = reqBody
```
BMartinos: I think we can just use 'largeBody' variable here and remove the variable declaration above

test/integration/transactionsAPITests.coffee
```diff
@@ -196,7 +207,119 @@ describe "API Integration Tests", ->
...
+              done err
+            else
+              # Then
+              Transaction.findOne { clientID: "999999999999999999999999" }, (error, newTransaction) ->
+                should.not.exist (error)
+                (newTransaction != null).should.be.true
+                newTransaction.routes[0].response.body.length.should.be.exactly utils.MAX_BODIES_SIZE
+                newTransaction.canRerun.should.be.true
+                done()
+
```
BMartinos: Maybe add a test to define what should happen when both request/response body has a large payload

test/integration/transactionsAPITests.coffee
```diff
@@ -325,6 +448,127 @@ describe "API Integration Tests", ->
...
+                  
+      it "should update transaction with large routes orchestrations request body", (done) ->
+        td = JSON.parse JSON.stringify transactionData
+        td.channelID = channel._id
+        clearTransactionBodies(td)
+        tx = new Transaction td
+        tx.save (err, result) ->
+          should.not.exist(err)
+          transactionId = result._id
+
```
BMartinos: dont think this space is needed here

src/utils.coffee
```diff
@@ -78,6 +78,8 @@ exports.serverTimezone = () ->
 # Use max 15 MiB leaving 1 MiB available for the transaction metadata
 mbs = config.api.maxBodiesSizeMB
 exports.MAX_BODIES_SIZE = MAX_BODIES_SIZE = if mbs <= 15 then mbs*1024*1024 else 15*1024*1024
+ta = config.api.truncateAppend
+tal = Buffer.byteLength ta
```
BMartinos: Maybe rename the above two variables to make it a bit more clearer what mean
```
taText
taLength
```

src/api/transactions.coffee
napiergit: Hi Martin, I don't see why this is necessary?

BMartinos: Is there any possibility that 'obj' variable could be something other than array/object? maybe a plain String or Boolean? and what would happen if it is

BMartinos:
Thanks @napiergit LGTM :+1:

PR #897, opened 2017-07-10
Add new config options

tmvumbi2:
Jira ticket: https://jembiprojects.jira.com/browse/OHM-448

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -2,7 +2,11 @@
   "bindAddress": "0.0.0.0",
   "mongo": {
     "url": "mongodb://localhost/openhim",
-    "atnaUrl": "mongodb://localhost/openhim"
+    "atnaUrl": "mongodb://localhost/openhim",
+    "connectionOptions": {},
+    "openHIMApiReadPreference": "primaryPreffered",
+    "openHIMApiWriteConcern": "1",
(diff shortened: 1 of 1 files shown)


config/default.json
```diff
@@ -2,7 +2,11 @@
   "bindAddress": "0.0.0.0",
   "mongo": {
     "url": "mongodb://localhost/openhim",
-    "atnaUrl": "mongodb://localhost/openhim"
+    "atnaUrl": "mongodb://localhost/openhim",
+    "connectionOptions": {},
+    "openHIMApiReadPreference": "primaryPreffered",
+    "openHIMApiWriteConcern": "1",
+    "openHIMApiReadConcern": "local"
```
BMartinos: Not something you added, but just noting it here, Mongoose doesnt seem to allow for `readConcern`

BMartinos: nvm i think it does, the docs points to certain options but this list thats present in mongodb should be allowed
http://mongoosejs.com/docs/connections.html#connection-string-options

BMartinos:
Thanks @tmvumbi2 ! The config looks good to me, Im just wondering if we should move the config options into the `connectionOptions` object, but lets leave it as is for now and can update it later

PR #898, opened 2017-07-10
Add function for mongo config

tmvumbi2:
Jira Ticket: https://jembiprojects.jira.com/browse/OHM-449

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -123,3 +123,13 @@ export function enforceMaxBodiesSize(ctx, tx) {
   ctx.totalBodyLength += len;
   return enforced;
 }
+
+export function getExtraMongoConfig() {
+  return {
+    db: {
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -123,3 +123,13 @@ export function enforceMaxBodiesSize(ctx, tx) {
   ctx.totalBodyLength += len;
   return enforced;
 }
+
+export function getExtraMongoConfig() {
```
BMartinos: Maybe rename this function to `getMongoOptions()` as this is only additional config options

tmvumbi: @BMartinos Done here: https://github.com/jembi/openhim-core-js/pull/899/commits/6d0f081337ff8118b0085a31600325983d2468a0

PR #899, opened 2017-07-11
Add options to mongo connection

tmvumbi:
Jira Ticket: https://jembiprojects.jira.com/browse/OHM-450

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -5,7 +5,7 @@
     "atnaUrl": "mongodb://localhost/openhim",
     "connectionOptions": {},
     "openHIMApiReadPreference": "primaryPreffered",
-    "openHIMApiWriteConcern": "1",
+    "openHIMApiWriteConcern": "majority",
     "openHIMApiReadConcern": "local"
   },
   "logger": {
--- a/config/default.json
+++ b/config/default.json
@@ -5,7 +5,7 @@
     "atnaUrl": "mongodb://localhost/openhim",
     "connectionOptions": {},
     "openHIMApiReadPreference": "primaryPreffered",
-    "openHIMApiWriteConcern": "majority",
+    "openHIMApiWriteConcern": 1,
     "openHIMApiReadConcern": "local"
   },
   "logger": {
--- a/src/api/certificateAuthority.js
+++ b/src/api/certificateAuthority.js
@@ -1,7 +1,7 @@
 import Q from "q";
 import logger from "winston";
 import pem from "pem";
-import { KeystoreModelAPI, CertificateAPI } from "../model/keystore";
+import { KeystoreModelAPI, CertificateModelAPI } from "../model/keystore";
 import * as utils from "../utils";
 import * as authorisation from "./authorisation";
 
--- a/src/api/certificateAuthority.js
+++ b/src/api/certificateAuthority.js
@@ -1,7 +1,7 @@
 import Q from "q";
 import logger from "winston";
 import pem from "pem";
-import { Keystore, Certificate } from "../model/keystore";
+import { KeystoreModelAPI, CertificateAPI } from "../model/keystore";
 import * as utils from "../utils";
 import * as authorisation from "./authorisation";
 
@@ -29,7 +29,7 @@ export function* generateCert() {
 }
 
 function* generateClientCert(options, ctx) {
-  const keystoreDoc = yield Keystore.findOne().exec();
+  const keystoreDoc = yield KeystoreModelAPI.findOne().exec();
 
     // Set additional options
   options.selfSigned = true;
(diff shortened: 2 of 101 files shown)


config/default.json
```diff
@@ -5,7 +5,7 @@
     "atnaUrl": "mongodb://localhost/openhim",
     "connectionOptions": {},
     "openHIMApiReadPreference": "primaryPreffered",
-    "openHIMApiWriteConcern": "1",
+    "openHIMApiWriteConcern": "majority",
```
BMartinos: I think this can be changed back to **1**, just ensure that it is a integer and not a string

BMartinos: I initially missed this as well:
https://docs.mongodb.com/manual/reference/write-concern/

BMartinos:
@tmvumbi as discussed in our chat earlier, just to spend some time researching how to update a mongoose connection with mongo options for specific request. 

The options should only be applied to API requests and not core requests, At the moment the model is defining the connection which is used in both core and API requests

src/api/metadata.js
```diff
@@ -2,21 +2,21 @@ import Q from "q";
...
-  Mediators: Mediator,
-  Users: User,
-  ContactGroups: ContactGroup,
-  Keystore
+  Channels: ChannelAPI,
+  Clients: ClientAPI,
+  Mediators: MediatorAPI,
+  Users: UserAPI,
+  ContactGroups: ContactGroupAPI,
+  KeystoreModelAPI
```
BMartinos: I think this file still has reference to the Keystore but not too sure how its being used...

src/model/alerts.js
```diff
@@ -1,4 +1,5 @@
 import { Schema } from "mongoose";
+import { connectionAPI } from "../config";
```
BMartinos: You can import multiple variables from the same file, lets add this on one line so that we dont have to duplicate the import of the config
https://github.com/jembi/openhim-core-js/pull/899/commits/38f9164829f0eb268b7e55ed12d879271af780ae#diff-9c101532f3045298b17eb1c108b13891R4

src/model/autoRetry.js
```diff
@@ -1,4 +1,5 @@
 import { Schema } from "mongoose";
+import { connectionAPI } from "../config";
```
BMartinos: Same as previous comment, Please apply to all models

src/model/keystore.js
```diff
@@ -28,5 +29,7 @@ const KeystoreSchema = new Schema({
 });
 
 // Model for storing the server key and cert as well as trusted certificates
+export const KeystoreModelAPI = connectionAPI.model("Keystore", KeystoreSchema);
```
BMartinos: rename this variable to `KeystoreAPI`

src/model/tasks.js
```diff
@@ -46,4 +47,6 @@ const TaskSchema = new Schema({
  * The task object that describes a specific task within the OpenHIM.
  * It provides some metadata describing a task and contains a number of transaction IDs.
  */
+export const TaskAPI = connectionAPI.model("Task", TaskSchema);
 export const Task = connectionDefault.model("Task", TaskSchema);
+
```
BMartinos: remove the extra line

src/api/certificateAuthority.js
```diff
@@ -1,7 +1,7 @@
 import Q from "q";
 import logger from "winston";
 import pem from "pem";
-import { Keystore, Certificate } from "../model/keystore";
+import { KeystoreModelAPI, CertificateAPI } from "../model/keystore";
```
tmvumbi2: @BMartinos I wasn't actually sure which name to use; in restart.js, there is a variable named 'KeystoreAPI' (`import * as KeystoreAPI from "../api/keystore";`), and the keystore model needs to be imported here as well (so I used a different name to avoid the conflict). Another possible solution could be use KeystoreAPI for the model, and rename the import reference, what do u think?

BMartinos: hmmmm i see... And having a look it it now i think the naming for that variable is more correct. Im wonder if we should rename these variables to be more description in terms of what they are, something like `KeystoreModelAPI`. This would indicate its a keystore model used for the API calls

BMartinos: So this one can stay as is, but the other ones will need to be updated to resemble the same format.

src/api/tasks.js
```diff
@@ -1,14 +1,14 @@
 
 import Q from "q";
 import logger from "winston";
-import { Task } from "../model/tasks";
-import { Transaction } from "../model/transactions";
-import { AutoRetry } from "../model/autoRetry";
+import { TaskModelAPI } from "../model/tasks";
+import { TransactionModelAPI } from "../model/transactions";
+import { AutoRetryModelAPI } from "../model/autoRetry";
 import * as Channels from "../model/channels";
```
tmvumbi2: @BMartinos This file is using an exported function 'isChannelEnabled' (line 98); we can import individually both ChannelModel and isChannelEnabled. I am not sure however if importing individual function is a pattern we use.

BMartinos: ok cool i think that is fine to keep as is then, I see its being used to determine if a channel is active as well and seems more readable in the way that is now

BMartinos:
Thanks @tmvumbi2 LGTM :+1:

PR #901, opened 2017-07-12
Ohm 447 fix symbol iterator bug

napiergit:


test/unit/transactions.js
```diff
@@ -0,0 +1,58 @@
...
+  })
+  
+  it("should calculate the bodies length of a transaction", async () => {
+    const lengthObj = { length: 0 }
+    let transaction = {}
+    try {
+      transactions.calculateTransactionBodiesByteLength(lengthObj, transaction, new WeakSet())
+      lengthObj.length.should.be.exactly(0)
+    } catch (e) {
+      e.should.not.exist()
```
BMartinos: `should.not.exist()` isnt a function of the `e` variable and throws and error. The assert should be `should.not.exist(e)`

test/unit/transactions.js
```diff
@@ -0,0 +1,58 @@
...
+  it("should update transaction metrics", async () => {
+    try {
+      const updates = {
+        $push: {
+          routes: {}
+        }
+      }
+      transactions.updateTransactionMetrics(updates, {})
+    } catch (e) {
+      e.should.not.exist()
```
BMartinos: `should.not.exist()` isnt a function of the `e` variable and throws and error. The assert should be `should.not.exist(e)`

PR #902, opened 2017-07-17
Ohm 445 fix memory leak

napiergit:


BMartinos:
screenshot from 2017-07-19 08-54-26

napiergit:
Yeah I also noticed that, but wasn't sure if it was an issue.  Seems pretty unfavourable though

PR #910, opened 2017-09-07
Update Mocha link

mortonfox:


rcrichton:
Thanks for the correction. We'll try get this on once another large restructuring branch is landed in master.

PR #912, opened 2017-09-11
OHM-464 Upgrade dependencies

bausmeier:


bausmeier:
The build is currently failing because the pinned pem module does not work with recent versions of OpenSSL. This will be addressed by OHM-487.

PR #916, opened 2017-10-06
Remove check for clientID and role name

tumijacob:
This commit removes a check for same role name and clientID. This check introduced a bug
and hence s need to removes it.


tumijacob:
The node versions are incompatible.

PR #923, opened 2017-10-24
OHM-516 Content is being re-encoded but original content-length header remains

bausmeier:


bausmeier:
Don't merge - a release needs to be tagged on this branch and then the conflicts need to be resolved. Please leave this to me to handle once the code has been reviewed.

PR #924, opened 2017-10-24
OHM-516 Content is being re-encoded but original content-length header remains

bausmeier:


bausmeier:
@nthfloor2 Looking in the OpenHIM transaction log it looks like the message has passed through fine, but the client actually receives a truncated response. See https://asciinema.org/a/OODdXstnezLsFCv7GpkQBeRYN which shows that the last few characters get cut off (the last word should be fizz). The actual content length is 73, but the gzipped content length is 71 in that case.

nthfloor2:
@bausmeier ok so using curl I see that it cuts off the last few characters - but still doing that on the branch with the new changes...

Interesting that my browser gives me the full response either way.

PR #931, opened 2017-11-10
Ohm 514 read only channel

andrew-w-ross:
Added methods to http channels that limit the methods that can be executed against it.


src/api/channels.js
```diff
@@ -207,6 +224,22 @@ export async function updateChannel (ctx, channelId) {
   // Set the user updating the channel for auditing purposes
   channelData.updatedBy = utils.selectAuditFields(ctx.authenticated)
 
+  if (!utils.isNullOrWhitespace(channelData.type) && utils.isNullOrEmpty(channelData.methods)) {
+    // Empty the methods if the type has changed from http
+    if (channelData.type !== 'http') {
+      channelData.methods = []
```
nthfloor2: Why are we only interested in checking the methods for http channels? Why not on the others?

PR #932, opened 2017-11-10
OHM-431 Username should be case IN-SENSITIVE on login

tumijacob:


src/api/users.js
```diff
@@ -24,7 +24,8 @@ export async function authenticate (ctx, email) {
   email = unescape(email)
 
   try {
-    const user = await UserModelAPI.findOne({email}).exec()
+    const emailReg = new RegExp(email, 'i')
```
andrew-w-ross: Escape the regular expression with [escapeRegExp](https://lodash.com/docs/4.17.4#escapeRegExp)

src/api/users.js
```diff
@@ -94,7 +95,7 @@ export async function userPasswordResetRequest (ctx, email) {
   }
 
   try {
-    const user = await UserModelAPI.findOneAndUpdate({email}, updateUserTokenExpiry).exec()
+    const user = await UserModelAPI.findOneAndUpdate({ email }, updateUserTokenExpiry).exec()
```
andrew-w-ross: If we are making the login case insensitive perhaps the same should be done for the password reset

src/api/users.js
```diff
@@ -252,7 +253,7 @@ export async function addUser (ctx) {
   }
 
   const userData = ctx.request.body
-
+  console.log(ctx.request.body)
```
andrew-w-ross: Remove please

src/api/users.js
```diff
@@ -261,8 +262,9 @@ export async function addUser (ctx) {
   userData.token = token
   userData.tokenType = 'newUser'
   userData.locked = true
+  userData.email = userData.email.toLocaleLowerCase()
```
andrew-w-ross: Why are we changing the locale? Use `toLowerCase()`

src/api/users.js
```diff
@@ -306,7 +308,7 @@ export async function getUser (ctx, email) {
   }
 
   try {
-    const result = await UserModelAPI.findOne({email}).exec()
+    const result = await UserModelAPI.findOne({ email }).exec()
```
andrew-w-ross: Make case insensitive

src/api/users.js
```diff
@@ -344,7 +346,7 @@ export async function updateUser (ctx, email) {
   if (userData._id) { delete userData._id }
 
   try {
-    await UserModelAPI.findOneAndUpdate({email}, userData).exec()
+    await UserModelAPI.findOneAndUpdate({ email }, userData).exec()
```
andrew-w-ross: Make case insensitive

src/api/users.js
```diff
@@ -24,7 +25,8 @@ export async function authenticate (ctx, email) {
   email = unescape(email)
 
   try {
-    const user = await UserModelAPI.findOne({email}).exec()
+    const emailReg = new RegExp(`^${_.escapeRegExp(email)}$`, 'i')
```
andrew-w-ross: Perhaps `new RegExp(`^${_.escapeRegExp(email)}$`, 'i')` should be moved into a separate function instead of being repeated.

src/api/users.js
```diff
@@ -94,7 +96,8 @@ export async function userPasswordResetRequest (ctx, email) {
   }
 
   try {
-    const user = await UserModelAPI.findOneAndUpdate({email}, updateUserTokenExpiry).exec()
+    const emailReg = new RegExp(`^${_.escapeRegExp(email)}$`, 'i')
```
andrew-w-ross: Extract into it's own function as mentioned above.

src/api/users.js
```diff
@@ -344,7 +348,8 @@ export async function updateUser (ctx, email) {
   if (userData._id) { delete userData._id }
 
   try {
-    await UserModelAPI.findOneAndUpdate({email}, userData).exec()
+    const emailReg = new RegExp(`^${_.escapeRegExp(email)}$`, 'i')
```
andrew-w-ross: Extract into it's own function as mentioned above.

src/api/users.js
```diff
@@ -368,7 +373,8 @@ export async function removeUser (ctx, email) {
   }
 
   try {
-    await UserModelAPI.findOneAndRemove({email}).exec()
+    const emailReg = new RegExp(`^${_.escapeRegExp(email)}$`, 'i')
```
andrew-w-ross: Extract into it's own function as mentioned above.

src/api/users.js
andrew-w-ross: Also this function now needs to be tested

src/api/users.js
andrew-w-ross: Also needs to be tested

src/api/users.js
andrew-w-ross: Also needs to be tested

PR #961, opened 2018-01-31
Update about.md

lindajembi:
Updating with new info from the Product Summary V4.0


BMartinos:
Codecov has dropped by 0.04 % which is why the build indicates its failing, only doc changes were made, and no code changes, merging into master

PR #966, opened 2018-02-21
Update about.md

lindajembi:
Added more text and additional diagram. Removed duplicate text.


docs/about.md
```diff
@@ -59,17 +59,44 @@ As the OpenHIM provides these centralised services it means that domain services
...
-* Minimal transaction overhead - The OpenHIM used the latest technologies such as Node.js and MongoDB to ensure that it doesn’t introduce any significant overhead to requests.
-
+## Components of the OpenHIM
+
+The OpenHIM logically consists of three components: 
+* The OpenHIM Core provides the main functions and services
+* The Administration Console provides an easy to use interface for system administrators to configure and manage the OpenHIM, giving a window into the workings of the HIE.
+* Mediators are additional services used to extend the functionality of the OpenHIM by transforming and orchestrating transactions.
+
+![OpenHIM Components](/_static/overview/OpenHIM Components.PNG)
```
BMartinos: We need to upload the image thats being referenced here. To do so, you can navigate to the `/docs/_static/overview`  directory (while in your banch) and click on the "**upload**" button, Just ensure that there are no spaces in the image name, these will need to be replaced with dashes

PR #975, opened 2018-03-23
OHM-648: fixed daily reports

nthfloor2:
Resolved unfinished promise which was failing silently

Replaced email-templates dependency with handlebars compiler.


package.json
```diff
@@ -52,11 +52,11 @@
     "chokidar": "^1.6.0",
     "co": "^4.1.0",
     "cookie": "^0.3.1",
-    "email-templates": "^3.0.6",
     "forever-monitor": "^1.5.2",
     "form-data": "^2.3.1",
+    "fs": "0.0.1-security",
```
rcrichton: `fs` is a standard module, you can remove this.

PR #986, opened 2018-11-09
Updated a metrics check that verifies the timestamp value to check that its a object (Date) instead of a number

BMartinos:
OHM-681

### DIFF
--- a/src/metrics.js
+++ b/src/metrics.js
@@ -25,7 +25,6 @@ export async function recordTransactionMetrics (transaction) {
       !transaction.response.timestamp || 
       !transaction.response.timestamp instanceof Date // Date object
   ) {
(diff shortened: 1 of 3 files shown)


src/metrics.js
```diff
@@ -20,7 +20,10 @@ async function recordTransactionMetric (fields, update) {
 }
 
 export async function recordTransactionMetrics (transaction) {
-  if (!transaction.response || typeof transaction.response.timestamp !== 'number') {
+  if (
+      !transaction.response || 
+      typeof transaction.response.timestamp !== 'object' // Date object
```
bausmeier: I think an `instanceof` check for `Date` would be better here if we know that it will be a date.

PR #989, opened 2018-11-19
OHM-651 channel audits body culling fix

Zooloo2014:


BMartinos:
Thanks @Zooloo2014 This is working great and the change looks good. Can we just add a test case to ensure those audits arent being returned in the API request
https://github.com/jembi/openhim-core-js/blob/master/test/integration/channelsAPITests.js#L754-L834

We can just add a few entries with the `lastBodyCleared` path when setting up the tests, And then verify that they arent being returned by the API request (IE remove them from the expectedPatches)

Zooloo2014:
The getChannelAudits integration is passing

PR #994, opened 2018-12-10
OHM-702 Report link fixes

bausmeier:


bausmeier:
Been losing my mind trying to debug these build failures today and it turns out that they only happen on Node.js v11.4.0. Upgrading to that version locally allows me to reliably reproduce the issue. I will continue debugging tomorrow.

PR #995, opened 2018-12-13
add notification settings to the travis yml file

bradsawadye:
This is so we can get slack notifications on travis builds

OHM-713 OHM-670

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,5 @@ before_install:
 after_success:
   - npm run coverage
   - ".travis/build_docker.sh"
- notifications:
-    slack: jembihealthsystems:mlQYVFbijxcZkesCt7G5VBoM
+notifications:
+  slack: jembihealthsystems:mlQYVFbijxcZkesCt7G5VBoM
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -16,3 +16,5 @@ before_install:
 after_success:
   - npm run coverage
   - ".travis/build_docker.sh"
+notifications:
+  slack: jembihealthsystems:mlQYVFbijxcZkesCt7G5VBoM
```
BMartinos: Lets add the on_success/on_failure settings to these notifications as well. 

Lets get notified everytime the build fails, but only get notified on success if the status has been updated.
This means that if the build is in a passing status and we add more commits that also pass the build, we dont get those notifications each time, only when the build fails will we be notified

https://docs.travis-ci.com/user/notifications#configuring-slack-notifications
```
notifications:
  slack:
    rooms:
      - <account>:<token>#development
      - <account>:<token>#general
    on_success: change # default: change
    on_failure: always # default: always
```

PR #997, opened 2019-01-02
Ohm 666 use calendar days for reports

bradsawadye:


README.md
```diff
@@ -23,6 +23,8 @@ Getting started with the OpenHIM-core
 
 To make use of your own custom configurations you can copy the [default.json](https://github.com/jembi/openhim-core-js/blob/master/config/default.json) config file and override the default setting:
 
+* Please specify the timezone's offset value otherwise the offset used will be that of the server's timezone.  You can do this by changing the value of the utcOffset in the default.json.
```
MattyJ007: You know what I mean

Also could you move it down to below the copy file instructions
```suggestion
 > To change the report bluh bluh bluh change the value of `utcOffset` in `config/default.json`.
```

README.md
```diff
@@ -23,14 +23,14 @@ Getting started with the OpenHIM-core
...
 
-* Please specify the timezone's offset value otherwise the offset used will be that of the server's timezone.  You can do this by changing the value of the utcOffset in the default.json.
-
 ```
 wget https://raw.githubusercontent.com/jembi/openhim-core-js/master/config/default.json
 # edit default.json, then
 openhim-core --conf=path/to/default.json
 ```
 
+> To specify the timezone that the openhim services are used in, change the utcOffset value in the default.json file. If the timezone's offset value is not specified, the dafault offset of +200 will be used. This ensures that weekly and daily channel reports have the correct timestamps.
```
MattyJ007: @bradsawadye Technically untrue - the default offset will only give the correct channel reports if the users are in that time zone :p (IE: Us). 
If we default the time to +2 we should say something along the lines of that the majority of known users are in South Africa.
But to be more generic +0 should be the default as I would think that would be the expected function.

Also, what happened with the discussion about the "moment timezone" package that would account for Daylight savings?

config/config.md
```diff
@@ -130,7 +130,9 @@ The following config option are provided by the OpenHIM. All of these options ha
   },
   "reports": {
     // To enable daily and weekly reporting
-    "enableReports": true
+    "enableReports": true,
+    //the default timezone offset
```
MattyJ007: @bradsawadye 
This Comment is not helpful.
Tell the user what they can do with this field.

config/config.md
```diff
@@ -130,7 +130,9 @@ The following config option are provided by the OpenHIM. All of these options ha
...
+    "utcOffset": +200
```
MattyJ007: @bradsawadye 
Is +200 correct?
Should it perhaps be "+0200" as it is used in the json config file. It may mislead the user

src/reports.js
```diff
@@ -11,6 +11,12 @@ import * as utils from './utils'
 
 config.reports = config.get('reports')
 
+let utcOffset = 0
```
MattyJ007: This appears to be an integer here when it is a String from the config. What does the moment package require?

MattyJ007: Maybe keep it consistent

src/reports.js
```diff
@@ -22,11 +28,11 @@ function sendReports (job, flag, done) {
   const channelMap = {}
 
   if (flag === 'dailyReport') {
-    from = moment().subtract(1, 'days').startOf('day').toDate()
-    to = moment().subtract(1, 'days').endOf('day').toDate()
+    from = moment().subtract(1, 'days').startOf('day').utcOffset(utcOffset).format("Do MMM YYYY HH:mm:ss Z")
```
MattyJ007: @bradsawadye Please could you rename these variables - they're a bit vague - to what? From who? The Variable should be self documenting

README.md
```diff
@@ -29,6 +29,8 @@ wget https://raw.githubusercontent.com/jembi/openhim-core-js/master/config/defau
...
```
MattyJ007: A bit clearer

```suggestion
> To specify the timezone in which the openhim services are used, change the utcOffset value in the default.json file. If the timezone's offset value is not specified, the dafault server time will be used. This ensures that weekly and daily channel reports have the expected timestamps. 
```

PR #1002, opened 2019-01-11
Ohm 732 fix flaky logs tests

BMartinos:


BMartinos:
The CodeCov fails due to the removal of the Log model Schema. This has reduced the amount of lines covered

PR #1005, opened 2019-01-16
Ohm 735 non admin api acl on get transactions

Zooloo2014:


BMartinos:
@Zooloo2014 i havent yet completed the code review/testing but i think we should also add a test to verify that a non-root user is able to retrieve the body when requesting multiple transactions

PR #1009, opened 2019-01-22
Ohm 737 release candidate version bump

BMartinos:


BMartinos:
@bausmeier yip, i removed the lock file to build it fresh with the new package version update, and any updated packages. Can revert this if this might cause issues.

bausmeier:
I don't think it will cause issues, but I think it should be mentioned in the changelog if we are updating dependencies.

PR #1012, opened 2019-02-05
will modify the failing tests

bradsawadye:
Some tests were failing on node version 10.15.1 This is a fix for the tests

OHM-751 OHM-748

### DIFF
--- a/test/integration/tcpIntegrationTests.js
+++ b/test/integration/tcpIntegrationTests.js
@@ -226,7 +226,6 @@ describe('TCP/TLS/MLLP Integration Tests', () => {
 
   afterEach(async () => {
     await Promise.all([
-      mockServer.close(),
       TransactionModel.deleteMany({})
     ])
     sandbox.reset()
@@ -245,6 +244,7 @@ describe('TCP/TLS/MLLP Integration Tests', () => {
 
     res.toString().should.eql(expectedResp)
     spy.callCount.should.eql(1)
+    mockServer.close()
   })
 
   it('will timeout a socket', async () => {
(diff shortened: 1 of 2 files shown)


test/unit/auditingTest.js
```diff
@@ -301,7 +301,7 @@ describe('Auditing', () => {
 
     after(async () => {
       config.auditing.auditEvents = JSON.parse(_restore)
-      await Promise.all(servers.map(s => s.close()))
+      await servers.map(s => s.close())
```
bausmeier: What is this line supposed to do now? Why the `.map` and `await`?

bradsawadye: Hey Brett, so I am still investigating, havent made much headway as to why they fail. The tests only pass when the close function is not in a promise. The program times out before the close is event is fired. I will push something that's a bit cleaner just now

test/integration/tcpIntegrationTests.js
```diff
@@ -228,6 +228,7 @@ describe('TCP/TLS/MLLP Integration Tests', () => {
     await Promise.all([
       TransactionModel.deleteMany({})
     ])
+    mockServer.close()
```
bausmeier: We're still not waiting for this to complete or handling any errors which might occur during close.

test/unit/auditingTest.js
```diff
@@ -301,7 +301,7 @@ describe('Auditing', () => {
 
     after(async () => {
       config.auditing.auditEvents = JSON.parse(_restore)
-      await servers.map(s => s.close())
+      servers.map(s => s.close())
```
bausmeier: We're still not waiting for the servers to close or handling any errors that may happen.

test/unit/auditingTest.js
bradsawadye: > What is this line supposed to do now? Why the `.map` and `await`?

the line is for closing the servers that are in an array

PR #1016, opened 2019-02-06
Ohm 751 update the read.me, the travis yml and the centos building script 

bradsawadye:


.travis.yml
```diff
@@ -1,7 +1,8 @@
 language: node_js
 node_js:
   - "lts/carbon"
-  - "lts/dubnium"
```
bausmeier: We should still build on the latest LTS, but just allow it to fail for now. We don't want to lose sight of this - we still need to resolve the issue ASAP.

packaging/build-docker-centos-rpm.sh
```diff
@@ -21,10 +21,10 @@ echo "Install needed packages: "
 docker exec -it $containerName sh -c "yum install -y git rpm-build redhat-rpm-config gcc-c++ make"
 
 echo "Install needed packages: "
-docker exec -it $containerName sh -c "curl --silent --location https://rpm.nodesource.com/setup_10.x | bash -"
+docker exec -it $containerName sh -c "curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash"
 
 echo "Install needed packages: "
-docker exec -it $containerName sh -c "yum install -y nodejs"
+docker exec -it $containerName sh -c "nvm install 10.15.0"
```
bausmeier: This pins the version for the build container which is correct, but we also need to pin the required Node.js version in the RPM package itself. See https://github.com/bbc/speculate#node-versions.

.travis.yml
```diff
@@ -1,13 +1,13 @@
 language: node_js
 node_js:
   - "lts/carbon"
-  - "10.13.0"
-  - "10.15.0"
```
bausmeier: I think that we should leave the build for 10.15.0 in as well at the moment since that's the highest version that we are supporting.

bausmeier:
@bradsawadye The RPM build is not working at the moment: `sh: nvm: command not found`. Please investigate and fix.

bausmeier:
@bradsawadye The RPM build was still failing because of an issue with the required nodejs version. I have fixed it now and made a few other small changes. Please will you review and test it.

bradsawadye:
The review of the new changes has been done and all looks good. The rpm package building and the package itself were tested, but using a version 5 release candidate. Testing of the package built on version 5 final release will be done when the release is done.

PR #1017, opened 2019-02-14
Support basic auth for API calls

bausmeier:
To simplify the authentication between other services and the OpenHIM
API.

### DIFF
--- a/src/api/authentication.js
+++ b/src/api/authentication.js
@@ -120,10 +120,13 @@ async function authenticateToken (ctx) {
 }
 
 async function authenticateRequest (ctx) {
-  // First attempt basic authentication
-  let user = await authenticateBasic(ctx)
-  if (user == null) {
-    // Otherwise try token based authentication
+  let user
+  // First attempt basic authentication if enabled
+  if (user == null && config.api.authenticationTypes.includes('basic')) {
+    user = await authenticateBasic(ctx)
+  }
+  // Otherwise try token based authentication if enabled
+  if (user == null && config.api.authenticationTypes.includes('token')) {
     user = await authenticateToken(ctx)
   }
   return user
--- a/src/api/authentication.js
+++ b/src/api/authentication.js
@@ -1,3 +1,4 @@
+import basicAuth from 'basic-auth'
 import crypto from 'crypto'
 import logger from 'winston'
 import atna from 'atna-audit'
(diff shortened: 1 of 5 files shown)


src/api/authentication.js
```diff
@@ -26,25 +27,51 @@ const auditingExemptPaths = [
 
 const isUndefOrEmpty = string => (string == null) || (string === '')
 
-export async function authenticate (ctx, next) {
+async function authenticateBasic (ctx) {
+  const credentials = basicAuth(ctx)
+  if (credentials == null) {
```
rcrichton: Is the use of `==` intentional? Automatic type casting doesn't read well and is generally not good style. Even though I don't see a problem with it for null checks.

src/api/authentication.js
```diff
@@ -56,57 +83,103 @@ export async function authenticate (ctx, next) {
...
+      `API request made by ${email} from ${ctx.request.host} has expired, denying access`,
+      { email }
+    )
   }
-  const user = await UserModelAPI.findOne({ email: caseInsensitiveRegex(email) })
-  ctx.authenticated = user
 
-  if (!user) {
+  const user = await UserModelAPI.findOne({ email: caseInsensitiveRegex(email) })
+  if (user == null) {
```
rcrichton: Looks like you made this change intentionally.

src/api/authentication.js
```diff
@@ -56,57 +83,103 @@ export async function authenticate (ctx, next) {
...
-      }
+  return user
+}
+
+async function authenticateRequest (ctx) {
+  // First attempt basic authentication
+  let user = await authenticateBasic(ctx)
+  if (user == null) {
+    // Otherwise try token based authentication
+    user = await authenticateToken(ctx)
```
rcrichton: I feel like we should maybe add a config option to enable both auth methods so one can be disabled in particular circumstances. Like if running over raw http or if some of the features (like replay prevention) from token auth are desired.

src/api/authentication.js
bausmeier: Yes it's intentional. I want to check for `null` or `undefined` but not other falsy values in this case.

src/api/authentication.js
bausmeier: The API is only available over HTTPS so prevention of replay attacks is handled by TLS. A config sounds like a good idea though. I will add one in and default to using the token style auth only so that the behaviour does not change.

bausmeier:
@rcrichton Thanks for the review. I'd like to add a simpler API token approach as well but in the interests of time basic auth meets our requirements for now. Having user credentials configured on all of the services that access the HIM and having to manage those is quite a pain at the moment. Tokens that can be easily issued and revoked would be great.

bausmeier:
@rcrichton I've made the authentication types configurable. Please take another look when you get a chance.

src/api/authentication.js
rcrichton: It would be better to be explicit them for readability, but not a big issue.

PR #1018, opened 2019-02-18
Support configuring authentication types as JSON

bausmeier:
After creating a release for the basic authentication support I realised that we need to be able to configure the authentication types using environment variables. This PR adds support for that.

It can be tested by starting the OpenHIM core with `api_authenticationTypes="[\"basic\"]" npm start`.

### DIFF
--- a/src/api/authentication.js
+++ b/src/api/authentication.js
@@ -119,15 +119,42 @@ async function authenticateToken (ctx) {
   return user
 }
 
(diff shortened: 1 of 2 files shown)


bausmeier:
Don't know what's up with the coverage report. It seems to think I've changed the `src/middleware/messageStore.js` file somehow.

bausmeier:
@rcrichton Any chance you have a few minutes to review this?

PR #1020, opened 2019-04-09
OHM-764 Add transformations to add/remove message body from transactions

Zooloo2014:
Prior to this change, message bodies were held in memory.

This change will persist message bodies to GridFS or retrieve the
bodies and recreate a complete in-memory transaction.

OHM-764 OHM-690


Zooloo2014:
Not needed

PR #1021, opened 2019-04-11
add a method for retrieving bodies

bradsawadye:
Transaction bodies are now stored in gridfs, and this commit adds the method for retrieving these bodies. Tests for this method have also been added

OHM-763


src/middleware/transactionBody.js
```diff
@@ -0,0 +1,26 @@
+import mongodb from 'mongodb'
+
+export const retrieveTransactionBody = async (db, fileId, callback) => {
```
bausmeier: @bradsawadye The function should be async *or* accept a callback - not both.

src/middleware/transactionBody.js
```diff
@@ -0,0 +1,26 @@
+import mongodb from 'mongodb'
+
+export const retrieveTransactionBody = (db, fileId, callback) => {
```
BMartinos: I dont think its nessecary to pass throught the db connection to the function. A connection already exists which we can re-use
https://github.com/jembi/openhim-core-js/pull/1019/files#diff-6b387023209fd79314641ba533f5ce23R3
https://github.com/jembi/openhim-core-js/pull/1019/files#diff-6b387023209fd79314641ba533f5ce23R48

src/middleware/transactionBody.js
```diff
@@ -0,0 +1,26 @@
...
+  if (!db) {
+      const err = new Error(`Transaction body retrieval failed. Database handle: ${db} is invalid`)
+      return callback(err, null)
+  }
+
+  if (!fileId) {
+    const err = new Error(`Transaction body retrieval failed: Transaction id: ${fileId}`)
```
BMartinos: This error isnt very descriptive as to why it failed. The description also makes reference to Transaction which it shouldnt be, as this function could be used for any GridFS document retrieval

src/middleware/transactionBody.js
```diff
@@ -0,0 +1,26 @@
+import mongodb from 'mongodb'
+
+export const retrieveTransactionBody = (db, fileId, callback) => {
+  if (!db) {
+      const err = new Error(`Transaction body retrieval failed. Database handle: ${db} is invalid`)
+      return callback(err, null)
```
BMartinos: We can remove the `null` here as we dont need to pass a success value

src/middleware/transactionBody.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  if (!fileId) {
+    const err = new Error(`Transaction body retrieval failed: Transaction id: ${fileId}`)
+    return callback(err, null)
```
BMartinos: We can remove the `null` here as we dont need to pass a success value

src/middleware/transactionBody.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  const bucket = new mongodb.GridFSBucket(db)
+
+  let body = ''
+  bucket.openDownloadStream(fileId)
+    .on('error', err => {
+      const error = new Error(`Transaction body retrieval failed: Error in reading stream: ${err.message}`)
```
BMartinos: The description also makes reference to Transaction which it shouldnt be, as this function could be used for any GridFS document retrieval

test/unit/transactionBodyTest.js
```diff
@@ -0,0 +1,97 @@
+/* eslint-env mocha */
+
+import { retrieveTransactionBody } from '../../src/middleware/transactionBody'
+import uuid from 'uuid'
+import mongodb from 'mongodb'
+import { config } from '../../src/config'
+import fs from 'fs'
```
BMartinos: This modules doesnt seem to be used for anything

test/unit/transactionBodyTest.js
```diff
@@ -0,0 +1,97 @@
...
+
+describe('retrieveTransactionBody()', async () => {
+  let db
+  let client
+  const mongoClient = mongodb.MongoClient
+
+  before(async () => {
+    client = await mongoClient.connect(config.mongo.url)
```
BMartinos: We dont have to create a new mongoose connection, but can instead use an existing mongoose connection to connect to.
https://github.com/jembi/openhim-core-js/pull/1019/files#diff-f8e8ecebde8954f0b9cebb0a9c5c9202R7

We can the define our before/after hooks for setting up/cleaning up the database connection
https://github.com/jembi/openhim-core-js/pull/1019/files#diff-f8e8ecebde8954f0b9cebb0a9c5c9202R11
https://github.com/jembi/openhim-core-js/pull/1019/files#diff-f8e8ecebde8954f0b9cebb0a9c5c9202R16

The for each test, we can clean out the collections to ensure a clean slate for the next text
https://github.com/jembi/openhim-core-js/pull/1019/files#diff-f8e8ecebde8954f0b9cebb0a9c5c9202R21

test/unit/transactionBodyTest.js
```diff
@@ -0,0 +1,97 @@
...
+    db = client.db()
+  })
+
+  after(() => {
```
bausmeier: @bradsawadye Your async test functions need to accept a callback parameter or return a promise. The current functions will complete before the callbacks have been called which may end up interfering with other tests and causing random failures like we are currently seeing on Travis. See https://mochajs.org/#asynchronous-code.

src/contentChunk.js
```diff
@@ -61,3 +61,29 @@ exports.extractStringPayloadIntoChunks = (payload) => {
     stream.end(payload)
   })
 }
+
+export const retrievePayload = (fileId, callback) => {
+  const db = connectionDefault.client.db()
+  if (!db) {
```
BMartinos: We dont need this logic to check the db. The connection is made in the connection.js file and only used here. Maybe what we can do is to create a common function for getting the GridFS bucket which can be used by both functions.

```
const getGridFSBucket = () => {
   return new mongodb.GridFSBucket(connectionDefault.client.db())
}
...
const bucket = getGridFSBucket()
```

src/contentChunk.js
```diff
@@ -61,3 +61,29 @@ exports.extractStringPayloadIntoChunks = (payload) => {
...
+      const err = new Error(`Payload retrieval failed. Database handle: ${db} is invalid`)
+      return callback(err)
+  }
+
+  if (!fileId) {
+    const err = new Error(`Payload retrieval failed: Payload id: ${fileId} is invalid`)
```
BMartinos: Lets update this description to state that "no fileId was supplied" as supplying a fileId of `somethingRandom12345` is considered invalid as its not an ObjectId.

src/contentChunk.js
```diff
@@ -61,3 +61,29 @@ exports.extractStringPayloadIntoChunks = (payload) => {
...
+    return callback(err)
+  }
+
+  const bucket = new mongodb.GridFSBucket(db)
+
+  let body = ''
+  bucket.openDownloadStream(fileId)
+    .on('error', err => {
+      const error = new Error(`Payload retrieval failed: Error in reading stream: ${err.message}`)
+      return callback(error)
```
BMartinos: We can just respond with the error that was thrown, instead of creating a new Error

test/unit/contentChunk.js
```diff
@@ -159,14 +160,72 @@ describe('contentChunk: ', () => {
...
+            body.should.eql(fileString)
+            done()
+          })
+        }
+      })
+
+      stream.end(fileString)
+    })
+
+    it('should return an error and null when file does not exist', (done) => {
```
BMartinos: The description of this test states that `null` is also returned which can be removed as we not testing for `null` and only care about the err object

src/contentChunk.js
```diff
@@ -61,3 +61,29 @@ exports.extractStringPayloadIntoChunks = (payload) => {
     stream.end(payload)
   })
 }
+
+export const retrievePayload = (fileId, callback) => {
```
BMartinos: To keep these functions consistent, can we update this method to be an async function instead of having a callback

test/unit/contentChunk.js
```diff
@@ -159,14 +160,72 @@ describe('contentChunk: ', () => {
...
+      const bucket = new mongodb.GridFSBucket(db)
+      const stream = bucket.openUploadStream()
+      const fileString = `JohnWick,BeowulfJohnWick,BeowulfJohnWick,BeowulfJohnWick,Beowulf
+                          JohnWick,BeowulfJohnWick,BeowulfJohnWick,BeowulfJohnWick,Beowulf
+                          JohnWick,BeowulfJohnWick,BeowulfJohnWick,BeowulfJohnWick,Beowulf
+                          JohnWick,BeowulfJohnWick,BeowulfJohnWick,BeowulfJohnWick,Beowulf
+                        `
+
+      stream.on('finish', (doc) => {
+        if(doc) {
```
bausmeier: @bradsawadye What happens if `doc` is falsy? Or if 'finish' is never emitted?

test/unit/contentChunk.js
```diff
@@ -14,48 +15,45 @@ describe('contentChunk: ', () => {
   });
 
   after(function() {
+    db.collection('fs.files').deleteMany({})
```
bausmeier: @bradsawadye All of the calls in this after function are asynchronous and need to be handled as such. We should await the completion of all of them.

bradsawadye: okay will sort it out

BMartinos:
@bradsawadye this branch seems to have gotten out of sync with master. Please pull master and resolve any conflicts that might exists.

bradsawadye:
The updating of transaction retrieving methods is on another branch. This is why the revert of commit https://github.com/jembi/openhim-core-js/pull/1021/commits/7960a0744f09b68edbe580deb754018984720a7f has been done above.

test/unit/contentChunk.js
```diff
@@ -159,14 +157,56 @@ describe('contentChunk: ', () => {
...
         should.deepEqual(result.length, payloadLength)
       })
     })
   })
+
+  describe('retrievePayload()', () => {
+    it('should return an error when the file id is null', async () => {
+        const fileId = null
+
+          retrievePayload(fileId).catch((err) => {
```
BMartinos: The spacing here seems inconsistent with the line above it

test/unit/contentChunk.js
```diff
@@ -13,49 +14,46 @@ describe('contentChunk: ', () => {
     db = client.db()
   });
 
-  after(function() {
-    MongoClient.close()
+  after(async () => {
+    await db.collection('fs.files').deleteMany({})
```
BMartinos: Im not sure moving the deletion to the after makes complete sense. This assumes that the database is in a clean state when starting the tests. It also means that the database doesnt get cleaned on a per test basic but rather as a test suite.

PR #1022, opened 2019-04-12
OHM-761 integrate content chunk when saving transaction

BMartinos:


src/utils.js
```diff
@@ -125,19 +125,19 @@ export const MAX_BODIES_SIZE = mbs >= 1 && mbs <= 15 ? mbs * 1024 * 1024 : 15 *
...
 
-export function enforceMaxBodiesSize (ctx, tx) {
+export function enforceMaxBodiesSize (ctx, body) {
   let enforced = false
 
   // running total for all bodies
   if ((ctx.totalBodyLength == null)) { ctx.totalBodyLength = 0 }
 
-  let len = Buffer.byteLength(tx.body)
+  let len = Buffer.byteLength(body)
```
Zooloo2014: Lots of the tests are failing here. This code doesn't cater for when the body is type string. Should there be a check of string/Array-/Buffer-type to decide how to get the length/byteLength correctly?

PR #1027, opened 2019-05-06
update the request and response in orchestations

bradsawadye:
The request and response bodies are stored separately in gridfs. The body  property has been replaced with the bodyId property, which is the id of the body stored in gridfs.

OHM-771


src/model/transactions.js
```diff
@@ -64,6 +64,19 @@ const ErrorDetailsDef = {
 
 // OrchestrationMetadata Schema
 const OrchestrationMetadataDef = {
+  name: {
```
BMartinos: Im not sure why we duplicating this schema? What we need to do is assign the `ResponseDefMain` (which will be updated later on) to the `OrchestrationMetadataDef.response`. And the same for the request property. That schema adds the victuals property `body` to the schema

bradsawadye: Its not being duplicated. Dont know how it came out like this. I added the RoutesOrchestrationsMetaDataDef, which you can see below the comment. I did this to get the tests involving the routes orchestrations to pass. The tests were expecting the old Orchestration schema which has a request and response body. This shall be changed when we work the routes.

BMartinos: I dont think its necessary to create the `RoutesOrchestrationMetadataDef` schema. All that we need to is assign the new schema as defined here https://github.com/jembi/openhim-core-js/blob/refactoring-features/src/model/transactions.js#L4-L36 to the orchestration schema definition here https://github.com/jembi/openhim-core-js/blob/refactoring-features/src/model/transactions.js#L71-L74

src/api/transactions.js
```diff
@@ -257,16 +257,44 @@ export async function getTransactions (ctx) {
...
-    const responseBodyChuckFileId = await extractStringPayloadIntoChunks(transaction.response.body)
+  if (transaction.response && transaction.response.hasOwnProperty('body')) {
+    if(transaction.response.body) {
+      transaction.response.bodyId = await extractStringPayloadIntoChunks(transaction.response.body)
+    }
     delete transaction.response.body
-    transaction.response.bodyId = responseBodyChuckFileId
+  }
+
+  if (transaction.orchestrations) {
```
MattyJ007: @bradsawadye This line can be combined with the line beneath

test/integration/transactionsAPITests.js
```diff
@@ -850,8 +850,8 @@ describe('API Integration Tests', () => {
         res.body[0].routes[0].request.body.should.equal(`<HTTP body${TRUNCATE_APPEND}`)
         res.body[0].routes[0].response.body.should.equal(`<HTTP resp${TRUNCATE_APPEND}`)
 
-        res.body[0].orchestrations[0].request.body.should.equal(`<HTTP body${TRUNCATE_APPEND}`)
-        res.body[0].orchestrations[0].response.body.should.equal(`<HTTP resp${TRUNCATE_APPEND}`)
+        // res.body[0].orchestrations[0].request.body.should.equal(`<HTTP body${TRUNCATE_APPEND}`)
```
MattyJ007: @bradsawadye What's happening here?

bradsawadye: The functionality for retrieving the orchestration bodies from gridfs has not been incorporated yet. Brent is working on that ticket. I've added a to do comment for the uncommenting once we integrate the grids functionality with the "orcs from Isengard"

src/contentChunk.js
```diff
@@ -132,3 +132,12 @@ const filterPayloadType = (transaction) => {
     resolve(transaction)
   })
 }
+
+export const addOrchestrationBodies = (transaction) => {
```
BMartinos: @Zooloo2014 im not sure if this function is needed. We already have a function to add the bodies to all the relevant parts for a transactions. We just need to incorporate the logic to add it still. Maybe we need to change that function name to be a bit more descriptive.
https://github.com/jembi/openhim-core-js/blob/refactoring-features/src/contentChunk.js#L114-L134

Zooloo2014: @BMartinos Wanted to check with you if this is the correct format for the orchestrations? 
https://github.com/jembi/openhim-core-js/blob/e8d050f54de4f81802156f2caf42b665ddc94b94/test/unit/contentChunk.js#L216-L233
https://github.com/jembi/openhim-core-js/blob/e8d050f54de4f81802156f2caf42b665ddc94b94/test/unit/contentChunk.js#L253-L257

Is the orchestrations array just a collection of transactions that can be mapped to the body retrieval function?

Zooloo2014: Ok, I see now... addOrchestrationBodies is actually the same as addBodiesToTransactions... I will fix this.

BMartinos: The orchestration request/response properties has the same schema as the main transaction request/response, so your definitions look fine.

src/api/transactions.js
```diff
@@ -246,6 +246,10 @@ export async function getTransactions (ctx) {
     // retrieve transaction request and response bodies
     const transformedTransactions = await addBodiesToTransactions(transactions)
 
+    if (transformedTransactions.orchestrations && transformedTransactions.orchestrations.length > 0) {
```
BMartinos: @Zooloo2014 I think we can actually move this logic into the `addBodiesToTransactions` function. So the `addBodiesToTransactions` function will have a transactions parameter for the entire transaction document (request/response/orchestrations/routes etc) and this function will resolve the payloads for all the transaction bodies

PR #1028, opened 2019-05-07
Ohm 690 code cleanup

BMartinos:
This PR deals with a few small cleanup tasks, but the one big task is the removal of the the max body size logic.

Because we are moving the transaction bodies into their own collection (GridFS) we no longer need to enforce the max allowed transaction size. This PR deals with removing the logic, as well as the associating tests


src/middleware/messageStore.js
```diff
@@ -131,6 +131,7 @@ export async function storeResponse (ctx, done) {
     delete update.response.body
     update.response.bodyId = responseBodyChuckFileId
   }
+  // console.log(update.response)
```
bradsawadye: remove console log

BMartinos: Addressed in 4aad9ac

PR #1029, opened 2019-05-10
Ohm 782 remove old transaction payloads

BMartinos:


src/contentChunk.js
```diff
@@ -85,13 +85,13 @@ const removeBodyById = (id) => {
   })
 }
 
-exports.promisesToRemoveAllTransactionBodies = (tx) => {
+export const promisesToRemoveAllTransactionBodies = (tx) => {
```
bradsawadye: This only removes the request and response bodies. Shouldn't we also remove the orchestration and route response and request bodies?

src/api/transactions.js
```diff
@@ -444,26 +444,31 @@ export async function updateTransaction (ctx, transactionId) {
...
-      const transaction = await TransactionModelAPI.findById(transactionId).exec()
       const channel = await ChannelModelAPI.findById(transaction.channelID).exec()
       if (!autoRetryUtils.reachedMaxAttempts(transaction, channel)) {
         updates.autoRetry = true
         await autoRetryUtils.queueForRetry(transaction)
       }
     }
 
+    // construct promises array to remove all old payloads
+    const removeBodyPromises = promisesToRemoveAllTransactionBodies(transaction)
```
bradsawadye: I think we should remove bodies based on whats been updated. If only the response is updated, we'll end up with no request body

src/contentChunk.js
BMartinos: Ive only implemented the removal of the main request/response body to get the basics functionality working while the orchestration/route ticket was still being worked on. Ill add in the logic for the others as well

src/api/transactions.js
BMartinos: Good point, ill update the function to analyse the update object and only remove the relevant bodies

src/bodyCull.js
```diff
@@ -40,11 +40,26 @@ async function clearTransactions (channel) {
     query['request.timestamp'].$gte = lastBodyCleared
   }
 
-  const transactionsToCullBody = await TransactionModel.find(query, { 'request.bodyId': 1, 'response.bodyId': 1, })
+  // constrcut promises array for removing transaction bodies
```
bradsawadye: typo on construct

PR #1030, opened 2019-05-10
OHM-776: Store route bodies in GridFS

bradsawadye:
The response and request body of the routes are stored in gridfs

OHM-777


BMartinos:
@bradsawadye As part of this ticket, we just need to do a bit of cleanup on the schema definitions. Due to working on these changes in separate tickets, we created a duplicated schema to keep the other functionality still working as it did, but we no longer need those duplicated schemas

src/contentChunk.js
```diff
@@ -135,6 +135,10 @@ export const addBodiesToTransactions = async (transactions) => {
       transaction.orchestrations = await addBodiesToTransactions(transaction.orchestrations)
     }
 
+    if (transaction.routes && transaction.routes.length > 0) {
```
bradsawadye: With the statement above, the code breaks if the transaction.routes is true but not an array. I think we should check if "transaction.routes" is an array

> _Array.isArray( transaction.routes)_ 

>

bradsawadye: Please also fix the condition for the orchestrations above

src/contentChunk.js
```diff
@@ -130,12 +140,16 @@ export const addBodiesToTransactions = async (transactions) => {
...
+  return Promise.all(transactions.map(async transaction => {
+    if (transaction.orchestrations &&
+        Array.isArray(transaction.orchestrations) &&
+        transaction.orchestrations.length > 0) {
       transaction.orchestrations = await addBodiesToTransactions(transaction.orchestrations)
     }
 
-    if (transaction.routes && transaction.routes.length > 0) {
+    if (transaction.routes &&
+        Array.isArray(transactions.routes) &&
```
bradsawadye: Typo, it should be transaction not transactions

PR #1031, opened 2019-05-17
OHM-778 Store routes bodies in GridFS

Zooloo2014:


Zooloo2014:
No longer required. 
We've moved work to merge on OHM-776-routes

PR #1032, opened 2019-05-21
add logic for removing outdated payloads

bradsawadye:
this is for the route bodies

OHM-783

### DIFF
--- a/src/contentChunk.js
+++ b/src/contentChunk.js
@@ -95,12 +95,34 @@ export const promisesToRemoveAllTransactionBodies = (tx) => {
       removeBodyPromises.push(() => removeBodyById(tx.response.bodyId))
     }
 
-    if (tx.orchestrations && tx.orchestrations.length > 0) {
-      for (let orch of tx.orchestrations) {
-        try {
-          removeBodyPromises = removeBodyPromises.concat(await promisesToRemoveAllTransactionBodies(orch))
-        } catch (err) {
-          return reject(err)
+    if (tx.orchestrations) {
+      if (typeof tx.orchestrations === 'object') {
+        removeBodyPromises = removeBodyPromises.concat(await promisesToRemoveAllTransactionBodies(tx.orchestrations))
+      }
+
+      if (Array.isArray(tx.orchestrations) && tx.orchestrations.length > 0) {
+        for (let orch of tx.orchestrations) {
+          try {
+            removeBodyPromises = removeBodyPromises.concat(await promisesToRemoveAllTransactionBodies(orch))
+          } catch (err) {
+            return reject(err)
+          }
+        }
+      }
+    }
+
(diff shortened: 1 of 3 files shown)


src/contentChunk.js
```diff
@@ -95,12 +95,34 @@ export const promisesToRemoveAllTransactionBodies = (tx) => {
...
     }
 
-    if (tx.orchestrations && tx.orchestrations.length > 0) {
-      for (let orch of tx.orchestrations) {
-        try {
-          removeBodyPromises = removeBodyPromises.concat(await promisesToRemoveAllTransactionBodies(orch))
-        } catch (err) {
-          return reject(err)
+    if (tx.orchestrations) {
+      if (typeof tx.orchestrations === 'object') {
```
BMartinos: Im not sure if this check is really needed here. Based on the transaction schema model, the orchestrations/routes will always be an array.
https://github.com/jembi/openhim-core-js/blob/master/src/model/transactions.js#L69-L70

The removal of transaction bodies are based on the record that is saved within the database which will always be an array for orchestrations/routes.

BMartinos: I think where some of the confusion might come in as with the `transactionUpdate` method which can do `$push` updates of the document in either an array, or an object format

test/unit/contentChunk.js
```diff
@@ -286,16 +286,27 @@ describe('contentChunk: ', () => {
 
     it('should return an array with promise functions to remove the payloads', async () => {
       const td = testUtils.clone(transaction)
+      td.orchestrations = [{ request: {}, response: {} }]
+      td.routes = [{ request: {}, response: {}, orchestrations: [{request: {}, response: {} }] }]
```
BMartinos: This is a little bit confusion to follow. Lets construct the object in one go to make it easier to read
```
..
td.orchestrations = [
  { 
    request: {
      bodyId: await testUtils.createGridFSPayload('<HTTP orchestration body request>')
    }, 
    response: {
      bodyId: await testUtils.createGridFSPayload('<HTTP orchestration body request>')
    } 
  }
]
...
```

test/unit/contentChunk.js
```diff
@@ -307,11 +318,21 @@ describe('contentChunk: ', () => {
       td.request.bodyId = requestBodyId
       td.response.bodyId = responseBodyId
 
+      td.orchestrations = [{ request: {}, response: {} }]
+      td.routes = [{ request: {}, response: {}, orchestrations: [{request: {}, response: {} }] }]
```
BMartinos: Same as the above comment about structuring the object a bit cleaner so that its easier to follow

src/contentChunk.js
bradsawadye: okay, will fix this now.

PR #1033, opened 2019-05-24
request and response streaming

bradsawadye:
OHM-784

### DIFF
--- a/src/koaMiddleware.js
+++ b/src/koaMiddleware.js
@@ -18,7 +18,8 @@ import * as pollingBypassAuthorisation from './middleware/pollingBypassAuthorisa
 import * as pollingBypassAuthentication from './middleware/pollingBypassAuthentication'
 import * as events from './middleware/events'
 import * as proxy from './middleware/proxy'
-import * as rewrite from './middleware/rewriteUrls'
+// TODO: uncomment the line below when url rewrititng is back in support
+//import * as rewrite from './middleware/rewriteUrls'
 import { config } from './config'
 
 config.authentication = config.get('authentication')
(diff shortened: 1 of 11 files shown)


src/koaMiddleware.js
```diff
@@ -18,7 +18,8 @@ import * as pollingBypassAuthorisation from './middleware/pollingBypassAuthorisa
 import * as pollingBypassAuthentication from './middleware/pollingBypassAuthentication'
 import * as events from './middleware/events'
 import * as proxy from './middleware/proxy'
-import * as rewrite from './middleware/rewriteUrls'
+// TODO: uncomment the line below when url rewrititng is back in support
```
BMartinos: Lets add the ticket number to which this TODO relates to. I think the ticket are already within Jira.
Also, Typo on `rewrititng`

src/koaMiddleware.js
```diff
@@ -67,7 +68,8 @@ export function setupApp (done) {
   app.use(messageStore.koaMiddleware)
 
   // URL rewriting middleware
-  app.use(rewrite.koaMiddleware)
+  // TODO: uncomment the code below when url rewriting is back in support
```
BMartinos: Same as above, regarding the TODO ticket number reference

src/middleware/rewriteUrls.js
```diff
@@ -1,161 +1,161 @@
...
-  // do nothing to the request
-  await next()
-  // on response rewrite urls
-  if (ctx.authorisedChannel.rewriteUrls) {
-    const rewrite = promisify(rewriteUrls)
-    ctx.response.body = await rewrite(ctx.response.body.toString(), ctx.authorisedChannel, ctx.authenticationType)
-    return winston.info(`Rewrote url in the response of transaction: ${ctx.transactionId}`)
-  }
-}
+// import url from 'url'
```
BMartinos: If we not calling this functionality of this script anywhere, we could just leave it as is. It doesnt need to be commented out as its not being called anywhere

src/middleware/router.js
```diff
@@ -1,4 +1,7 @@
-import zlib from 'zlib'
+// All the gzip functionality is being commented out
+// TODO: uncomment the gzip functions
```
BMartinos: same as above, just add a ticket number to this TODO as a reference

PR #1034, opened 2019-05-27
OHM-785 receive request stream

Zooloo2014:
PR to investigate and add streaming to GridFS and downstream, and updating transaction after response

### DIFF
--- a/src/middleware/router.js
+++ b/src/middleware/router.js
@@ -1,7 +1,7 @@
 // All the gzip functionality is being commented out
 // TODO: OHM-693 uncomment the gzip functions when working on ticket
 
-import zlib from 'zlib'
+//import zlib from 'zlib'
 import http from 'http'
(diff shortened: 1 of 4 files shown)


src/middleware/router.js
```diff
@@ -215,16 +215,17 @@ function sendRequestToRoutes (ctx, routes, next) {
                 ctx.autoRetry = true
                 ctx.error = responseObj.error
               }
-
               // then set koa response from responseObj.response
-              return setKoaResponse(ctx, responseObj.response)
+              /* return */ setKoaResponse(ctx, responseObj.response)
             } else {
-              return setKoaResponse(ctx, response)
+              /* return */ setKoaResponse(ctx, response)
```
BMartinos: Lets maybe remove the commented out `return`, unless this is just commented out temporarily for now and needs to be added back in later on.

PR #1035, opened 2019-06-04
Ohm 790 store response middleware

bradsawadye:


src/middleware/messageStore.js
```diff
@@ -174,19 +190,49 @@ export async function storeResponse (ctx, done) {
...
+      logger.error(`Could not save completed response metadata for transaction: ${ctx.transactionId}. ${err}`)
+      return done(err) 
     }
     if ((tx === undefined) || (tx === null)) {
       logger.error(`Could not find transaction: ${ctx.transactionId}`)
       return done(err)
     }
-    logger.info(`stored primary response for ${tx._id}`)
-    return done()
+    logger.info(`stored completed primary response for ${tx._id}`)
```
BMartinos: I think we can remove the text `completed` as this might be interpreted as the transaction status (Successful, Completed, Failed etc)

src/middleware/router.js
```diff
@@ -13,10 +13,19 @@ import * as utils from '../utils'
...
+import { Writable, Readable } from 'stream';
+import util from 'util'
+import mongodb from 'mongodb'
+import { connectionDefault } from '../config'
+import { brotliCompressSync } from 'zlib';
 
 config.mongo = config.get('mongo')
 config.router = config.get('router')
 
+
```
BMartinos: Remove this unnecessary space

src/middleware/router.js
```diff
@@ -13,10 +13,19 @@ import * as utils from '../utils'
 import * as messageStore from '../middleware/messageStore'
 import * as events from '../middleware/events'
 import { promisify } from 'util'
+import { getGridFSBucket } from '../contentChunk'
+import { Writable, Readable } from 'stream';
+import util from 'util'
+import mongodb from 'mongodb'
+import { connectionDefault } from '../config'
+import { brotliCompressSync } from 'zlib';
```
BMartinos: Not sure what this is used for? We shouldn't include modules and functions that arent being used

src/middleware/router.js
```diff
@@ -13,10 +13,19 @@ import * as utils from '../utils'
 import * as messageStore from '../middleware/messageStore'
 import * as events from '../middleware/events'
 import { promisify } from 'util'
+import { getGridFSBucket } from '../contentChunk'
+import { Writable, Readable } from 'stream';
+import util from 'util'
+import mongodb from 'mongodb'
```
BMartinos: This doesnt seems to be used anywhere within the code.

src/middleware/router.js
```diff
@@ -13,10 +13,19 @@ import * as utils from '../utils'
...
+import { connectionDefault } from '../config'
```
BMartinos: This doesnt seems to be used anywhere within the code.

src/middleware/router.js
```diff
@@ -13,10 +13,19 @@ import * as utils from '../utils'
 import * as messageStore from '../middleware/messageStore'
 import * as events from '../middleware/events'
 import { promisify } from 'util'
+import { getGridFSBucket } from '../contentChunk'
+import { Writable, Readable } from 'stream';
+import util from 'util'
```
BMartinos: This doesnt seems to be used anywhere within the code.

src/middleware/router.js
```diff
@@ -432,8 +430,13 @@ function sendHttpRequest (ctx, route, options) {
     const routeReq = method.request(options, (routeRes) => {
       response.status = routeRes.statusCode
       response.headers = routeRes.headers
-      response.body = []
+      response.body = new Readable()
+      response.body._read = () => {}
```
BMartinos: What does this `_read ` property do? And why do we need to attach an empty function to it?

src/middleware/router.js
```diff
@@ -450,16 +453,69 @@ function sendHttpRequest (ctx, route, options) {
...
+        }).on('error', (err) => {
+            logger.error('Error sending Response upstream: ' + JSON.stringify(err))
+            reject(err)
+          })
+          .on('finish', () => {
+            logger.info(`Finished sending Response upstream`)
+            resolve(response)
+          })
+
+        uploadStream = bucket.openUploadStream()
```
BMartinos: Could we not maybe re-use the contentChunk function already exists, instead of writing all of this again. If it functions differently then we might need to re-think /refactor the contentChunk functions

src/middleware/messageStore.js
```diff
@@ -174,19 +190,49 @@ export async function storeResponse (ctx, done) {
     update.orchestrations.push(...ctx.orchestrations)
   }
 
-  await extractTransactionPayloadIntoChunks(update)
+  //await extractTransactionPayloadIntoChunks(update)
```
BMartinos: Should this be removed instead of just commented out?

src/middleware/router.js
```diff
@@ -450,16 +453,69 @@ function sendHttpRequest (ctx, route, options) {
...
+          })
+      }
 
       // See https://www.exratione.com/2014/07/nodejs-handling-uncertain-http-response-compression/
       routeRes
         .on('data', (chunk) => {
-          bufs.push(chunk)
+          if (!ctx.responseTimestamp) {
+            ctx.responseTimestamp = new Date()
+            messageStore.initiateResponse(ctx, () => {})
```
BMartinos: We dont seem to be handling the callback for this function. The callback receives an error and transaction variable depending on when the callback was triggered. We either need to handle the callback response, or not have a callback at all

src/middleware/router.js
```diff
@@ -450,16 +453,69 @@ function sendHttpRequest (ctx, route, options) {
...
+          }
+          uploadStream.write(chunk)
+          ctx.response.body.write(chunk)
+          response.body.push(chunk)
+          //bufs.push(chunk)
```
BMartinos: Can this be removed?

src/middleware/router.js
```diff
@@ -450,16 +453,69 @@ function sendHttpRequest (ctx, route, options) {
...
       routeRes
         .on('data', (chunk) => {
-          bufs.push(chunk)
+          if (!ctx.responseTimestamp) {
+            ctx.responseTimestamp = new Date()
+            messageStore.initiateResponse(ctx, () => {})
+          }
+          uploadStream.write(chunk)
+          ctx.response.body.write(chunk)
+          response.body.push(chunk)
```
BMartinos: What is the difference between `ctx.response.body.write` and `response.body.push` here?

src/middleware/router.js
```diff
@@ -553,11 +610,37 @@ function sendHttpRequest_OLD (ctx, route, options) {
...
+          response.startTimestamp = new Date()
+        }
+        uploadStream.write(chunk)
+        bufs.push(chunk)
+      })
 
       // See https://www.exratione.com/2014/07/nodejs-handling-uncertain-http-response-compression/
       routeRes.on('end', () => {
         response.timestamp = new Date()
+        response.endTimestamp = new Date()
```
BMartinos: This property doesnt seem to be used anywhere within the codebase?

src/middleware/router.js
```diff
@@ -776,6 +859,6 @@ function isMethodAllowed (ctx, channel) {
 export async function koaMiddleware (ctx, next) {
   const _route = promisify(route)
   await _route(ctx)
-  await messageStore.storeResponse(ctx, () => {})
+  //await messageStore.storeResponse(ctx, () => {})
```
BMartinos: Should this still be commented out? Are we saving the response elsewhere or does this need to come back in at some point?

test/unit/routerTest.js
```diff
@@ -63,6 +63,17 @@ describe('HTTP Router', () => {
         ctx.response.header.should.be.ok
       })
 
+      it('should route an incomming http request and then stream the response into gridfs', async () => {
```
BMartinos: Typo: `incomming`

src/contentChunk.js
```diff
@@ -3,7 +3,7 @@ import mongodb from 'mongodb'
 import { connectionDefault } from './config'
 
 let bucket
-const getGridFSBucket = () => {
+export const getGridFSBucket = () => {
```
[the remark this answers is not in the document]
bradsawadye: I exported it so it can be used in the router and in the koamiddleware (rawBodyReader f(x)) where bucket creation is done. And this will also ensure that we have one bucket instance being used throughout. The insert function cannot be used because it takes the whole payload as an arg. We are getting chunks and streaming them into gridfs.

src/middleware/router.js
bradsawadye: Isn't we are going to have a start timestamp (when the stream starts) and an end timestamp (when the stream ends)?

src/middleware/router.js
Zooloo2014: The comment can be removed. 
Without the 'messageStore.storeResponse' call here, the rerunApp, tcpApp and pollingApp functions won't be able to update transactions until we refactor them.

src/middleware/router.js
bradsawadye: I'm not sure about this, but it seems the 'ctx.response.body.write(chunk)' doesn't work. From the tests I've done, the 'response.body.push' is what works so I'm going to strip out the 'ctx.response.body.write(chunk)'. The function setKoaresponse sets the ctx.response based on the response.

src/middleware/router.js
```diff
@@ -13,10 +13,14 @@ import * as utils from '../utils'
 import * as messageStore from '../middleware/messageStore'
 import * as events from '../middleware/events'
 import { promisify } from 'util'
+import { getGridFSBucket } from '../contentChunk'
 
 config.mongo = config.get('mongo')
 config.router = config.get('router')
 
+
+var bucket
```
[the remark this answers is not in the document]
bradsawadye: addressed

src/middleware/messageStore.js
```diff
@@ -90,9 +88,10 @@ export async function initiateRequest (ctx) {
 
     tx.save((err, tx) => {
       if (err) {
-        logger.error(`Could not save transaction metadata: ${err}`)
+        logger.error(`Could not save transaction metadata (intial request): ${err}`)
```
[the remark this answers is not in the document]
bradsawadye: addressed

src/middleware/messageStore.js
bradsawadye: addressed

src/middleware/router.js
bradsawadye: addressed

src/middleware/router.js
bradsawadye: fixed

test/unit/routerTest.js
bradsawadye: fixed

src/middleware/router.js
BMartinos: Reading up on this a bit, the `.write` doesnt seem valid and should be removed. The koa `ctx.body` / `ctx.response.body` can accept chucks being pushed to it, or it can accept an entire stream directly piped into it. In our case we need to push the chucks as we are using the chucks for other things as well. E.g `initiateResponse`

src/middleware/router.js
BMartinos: Sure, but when doing a search for that property it only appears where its defined, so its currently not sued anywhere. I also think the naming of the property elsewhere has been set to something like `timestampEnd` instead of `endTimestamp`

src/middleware/router.js
BMartinos: Ok cool so for the new changes we are doing this isnt needed. but for the rerun/tcp/polling apps this will still be needed (until we refactor).

Lets maybe just add a comment here as to why this is commented out, and that we need to address this for the other applications served via Koa

src/contentChunk.js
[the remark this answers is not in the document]
BMartinos: Ok cool that makes sense. Im just not a fan of having a module that does this for us, and then we implement similar code elsewhere. Maybe we should look at refactoring the GridFS stuff a bit more to better align with this work

src/middleware/router.js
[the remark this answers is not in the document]
bradsawadye: commit that fixes this - https://github.com/jembi/openhim-core-js/pull/1036/commits/7afbbafb09eb82730888261e1ef66c9228120fc6

src/middleware/messageStore.js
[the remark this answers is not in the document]
bradsawadye: commit that fixes this - https://github.com/jembi/openhim-core-js/pull/1036/commits/7afbbafb09eb82730888261e1ef66c9228120fc6

src/middleware/messageStore.js
bradsawadye: commit fixing this - https://github.com/jembi/openhim-core-js/pull/1036/commits/7afbbafb09eb82730888261e1ef66c9228120fc6

src/middleware/router.js
bradsawadye: commit fixing this - https://github.com/jembi/openhim-core-js/pull/1036/commits/1d490191fae6ec448f38737d313ce667923f18c4

src/middleware/router.js
bradsawadye: Its already removed on the new branch

test/unit/routerTest.js
bradsawadye: commit fixing this - https://github.com/jembi/openhim-core-js/pull/1036/commits/7afbbafb09eb82730888261e1ef66c9228120fc6

BMartinos:
Closing this PR as this code has been merged into https://github.com/jembi/openhim-core-js/pull/1036

PR #1036, opened 2019-06-05
OHM-692 merge request response

Zooloo2014:
This branch merges the code of the request streaming and response streaming into a single source


BMartinos:
@Zooloo2014 For ticket OHM-791, The transaction response seems to be overriden once the final response is updated. When i have a look at the openhim-console dashboard, theres a bunch of missing information on the transaction. This seems to be removed when an update of the transactions happens

PR #1037, opened 2019-06-14
create a function for sending an http request

bradsawadye:
This is the function that will be used for sending secondary routes requests. It uses some of the functionality that is used in the function for sending the primary route requests. Refactoring will be done later on to apply the DRY principle. The function has been added to the 'sendRequest' function, which is the wrapper for the functions that send the requests. A flag has been added to use the function for secondary routes when route is secondary.

OHM-800

### DIFF
--- a/src/middleware/router.js
+++ b/src/middleware/router.js
@@ -526,6 +526,7 @@ function sendHttpRequest (ctx, route, options) {
       })
       .on('error', (err) => {
         ctx.primaryRouteFailure = true
+        messageStore.initiateResponse(ctx, () => {})
         logger.error(`Error streaming response upstream: ${err}`)
         reject(err)
       })
--- a/src/middleware/router.js
+++ b/src/middleware/router.js
@@ -281,7 +281,7 @@ function sendRequestToRoutes (ctx, routes, next) {
 
     Promise.all(promises).then(() => {
       logger.info(`All routes completed for transaction: ${ctx.transactionId}`)
-      
+
       // Set the final status of the transaction
       setTransactionFinalStatus(ctx)
 
@@ -452,6 +452,12 @@ function sendHttpRequest (ctx, route, options) {
       method = https
     }
 
+    /*
+     This is used to determine whether the secondary route should finish execution.
+     It should not when there is a failure on the primary route
+    */
+    ctx.primaryRouteFailure = false
+
     const routeReq = method.request(options)
       .on('response', (routeRes) => {
         response.status = routeRes.statusCode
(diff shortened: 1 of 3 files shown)


src/middleware/router.js
```diff
@@ -281,16 +281,10 @@ function sendRequestToRoutes (ctx, routes, next) {
...
-        if (err) {
-          logger.error(`Setting final status failed for transaction: ${ctx.transactionId}`, err)
-          return
-        }
-        logger.debug(`Set final status for transaction: ${ctx.transactionId}`)
-      })
- */
+      setTransactionFinalStatus(ctx)
+
       // TODO: OHM-694 Uncomment when secondary routes are supported
```
BMartinos: I think we need to also look at bringing this code back in now that the secondary routes are back in again

src/middleware/router.js
```diff
@@ -441,6 +452,12 @@ function sendHttpRequest (ctx, route, options) {
       method = https
     }
 
+    /*
+     This is used to determine whether the secondary route should finish execution.
+     It should not when there is a failure on the primary route
```
BMartinos: Reading this comment on its own seems a bit ambiguous in terms of what its trying to address

src/middleware/router.js
```diff
@@ -498,7 +515,7 @@ function sendHttpRequest (ctx, route, options) {
         ctx.res.socket
           .on('finish', () => {
             messageStore.completeResponse(ctx, (err, tx) => {
-              setTransactionFinalStatus(ctx)
+              // setTransactionFinalStatus(ctx)
```
BMartinos: Lets actually remove this `.on('finish')` handler as its not doing anything

src/middleware/router.js
```diff
@@ -537,6 +556,110 @@ function sendHttpRequest (ctx, route, options) {
...
+  return new Promise((resolve, reject) => {
+    const response = {}
+    let { downstream } = ctx.state
+    let method = http
+
+    if (route.secured) {
+      method = https
+    }
+
+    const routeReq = method.request(options)
```
BMartinos: Is it possible to refactor the methods for sending out the stream to be re-usable. Alot of this code seems to be duplicated on the sending of the primary request

src/middleware/router.js
```diff
@@ -537,6 +556,110 @@ function sendHttpRequest (ctx, route, options) {
...
+        const responseBuf = []
+        routeRes
+          .on('data', chunk => {
+            if (!response.timestamp) {
+              response.timestamp = new Date()
+            }
+
+            // Storing of the secondary route response is not done when the primary route fails
+            if (ctx.primaryRouteFailure) {
+              uploadStream.abort(() => {
```
BMartinos: Im not sure this is working as expected. If the main route is isnt reachable, i can still see the secondary route being sent while the transactions has failed

bradsawadye: Fixed, now working

PR #1038, opened 2019-06-19
OHM-807 Streaming changes for rerun transactions

Zooloo2014:


src/middleware/messageStore.js
```diff
@@ -357,13 +389,12 @@ export function setFinalStatus (ctx, callback) {
       logger.debug(`The transaction status has been set to ${ctx.mediatorResponse.status} by the mediator`)
       update.status = ctx.mediatorResponse.status
     } else {
-      tx.status = getContextResult()
+      //tx.status = getContextResult()
```
BMartinos: This commented out code can be removed

src/tasks.js
```diff
@@ -136,6 +136,11 @@ function rerunTransaction (transactionID, taskID, callback) {
   rerunGetTransaction(transactionID, (err, transaction) => {
     if (err) { return callback(err) }
 
+    if (['POST', 'PUT'].includes(transaction.request.method) && (!transaction.request.bodyId)) {
```
BMartinos: I think we should include the PATCH method as well

src/tasks.js
```diff
@@ -225,6 +221,13 @@ function rerunSetHTTPRequestOptions (transaction, taskID, callback) {
   options.headers.parentID = transaction._id
   options.headers.taskID = taskID
 
+  /*
+   *  For GET and DELETE, bodyId will be null. Still need to supply
+   *     empty header, so that HIM will not expect a body in GridFS
+   *  For POST and PUT, bodyId will be fileId for body stored in GridFS
+   */
+  options.headers['X-Body-ID'] = transaction.request.bodyId
```
BMartinos: For consistency, lets keep the cases the same for these headers
X-Body-ID | x-body-id

src/tasks.js
```diff
@@ -260,8 +263,11 @@ function rerunHttpRequestSend (options, transaction, callback) {
   logger.info(`Rerun Transaction #${transaction._id} - HTTP Request is being sent...`)
   const req = http.request(options, (res) => {
     res.on('data', chunk => {
-      // response data
-      response.body += chunk
+      /*
+       *  Don't need the response body at this point, because it's already been captured 
+       *  in GridFS (from router.js). 
+       *  Still need to have 'data' listener defined, or it changes program behaviour
+       */
```
BMartinos: Im not sure im fully understanding this. Why arent we doing anything with this request data coming back? or rather, why is it being processed here and somewhere else so that we dont need the response here

Zooloo2014: The rerun task is a wrapper for a routed transaction. The routed transaction is always stored in GridFS. But for reruns, there is an extra (higher) layer where the task runner submits a request to the rerun port, and gets a response back on the rerun port. What this area of the code is capturing is the response from the rerun port, which at this point has already been sent to gridFS as the routed transaction response. This response is discarded further up the call stack anyway (in the non-streaming version of the core); we're just not accumulating all the chunks here and taking the memory hit.

BMartinos: ok great that makes sense. So this is the response from the rerun request (Request sent to rerun port).

Zooloo2014:
Manually merged into OHM-692-merge-request-response

PR #1040, opened 2019-06-27
5.3.0 alpha.1

BMartinos:
This Pull Request is for the Alpha release for the OpenHIM core routing refactoring changes that has been implemented.

The routing refactoring changes made various changes to how the routing engine works within the OpenHIM. Previously, all payload messages were kept in memory while routing the transaction through to its destinations. While for small transactions this is fine, when bigger transactions are processed the OpenHIM can run out of memory and cause performance issues.

The OpenHIM also had to wait for the entire message to be read into memory before continuing the routing process, and the same applies for the response. They would need to be processed completely before responding to the client. 

To combat these issues, we are not reading the entire incoming/outgoing payload into memory, but instead streaming it directly into MongoDB GridFS. This also allows us to remove the body truncate limit for storing payloads as the MongoDB limit of 16mb no longer applies.

Note: Some features have been temporarily dropped while developing this refactoring changes and will be coming back in in future, proved there is still a need for them.


src/middleware/messageStore.js
```diff
@@ -26,132 +26,280 @@ function copyMapWithEscapedReservedCharacters (map) {
...
     }
-  }
 
-  if (utils.enforceMaxBodiesSize(ctx, tx.request)) { tx.canRerun = false }
+    /*
+     *  For short transactions, the 'end' timestamp is before the 'start' timestamp.
+     *  (Persisting the transaction initially takes longer than fully processing the transaction)
+     *  In these cases, swop the 'start' and 'end' values around; the transaction duration is
+     *  not exactly accurate, but at least it isn't negative.
+     */
```
rcrichton: This probably means that we aren't capturing the start timestamp in the right place. Rather than this hack we should try find a reliable place to capture the start timestamp.

src/middleware/messageStore.js
```diff
@@ -26,132 +26,280 @@ function copyMapWithEscapedReservedCharacters (map) {
...
+
+  const headers = copyMapWithEscapedReservedCharacters(ctx.response.header)
+/*
   // check if channel response body is false and remove
   if (ctx.authorisedChannel.responseBody === false) {
     // reset request body - primary route
     res.body = ''
   }
-
+ */
```
rcrichton: Is this no longer necessary? If not it should be removed.

src/middleware/messageStore.js
```diff
@@ -26,132 +26,280 @@ function copyMapWithEscapedReservedCharacters (map) {
...
-    orchestrations: []
+    'response.status': ctx.response.status,
+    'response.headers': headers,
+    'response.bodyId': ctx.response.bodyId,
+    'response.timestamp': ctx.responseTimestamp,
+    error: ctx.error
   }
 
-  utils.enforceMaxBodiesSize(ctx, update.response)
+  //await extractTransactionPayloadIntoChunks(update)
```
rcrichton: Is this no longer necessary? If not it should be removed.

src/middleware/streamingReceiver.js
```diff
@@ -0,0 +1,364 @@
...
+      size += chunk.toString().length
+      logger.info(`Read request CHUNK # ${counter} [ Total size ${size}]`)
+
+      // Write chunk to GridFS & downstream
+      if (storeRequestBody && !bodyId) {
+        gridFsStream.write(chunk)
+      }
+
+      ctx.state.downstream.push(chunk)
+    })
```
rcrichton: Reading all the data in chunk out of a stream and into another stream doesn't seem necessary, we could just us `.pipe()` to pipe this into the gridFsStream and then assign the `ctx.req` stream to `ctx.state.downstream` rather than copy chunk over one at a time.

src/middleware/streamingReceiver.js
```diff
@@ -0,0 +1,364 @@
...
+      .on('data', (chunk) => {
+        if (allowRequest) {
+          counter++;
+          size += chunk.toString().length
+          logger.info(`Read request CHUNK # ${counter} [ Total size ${size}]`)
+
+          bodyCopy.push(chunk)
+          ctx.state.downstream.push(chunk)
+        }
+      })
```
rcrichton: As above

src/middleware/streamingRouter.js
```diff
@@ -0,0 +1,269 @@
...
+    downstream
+      .on('data', (chunk) => {
+        if (options.requestBodyRequired) {
+          routeReq.write(chunk)
+        }
+        if (!startedRequest && statusEvents.startRequest) {
+          statusEvents.startRequest()
+          startedRequest = true
+        }
+      })
```
rcrichton: Here we could again `.pipe()` downstream to `routeReq` we could probably start the request as soon as we start piping instead of on the first chunk and then checking this every time.

PR #1041, opened 2019-06-27
Bug fix: Responses in Mediator format was failing the transaction

BMartinos:
This was due to the change to read an incoming stream instead of a full payload.
This change takes the reponse.body stream and builds up a string with all the supplied chuncks before setting the koa response.
We need to read in the entire payload here to process the mediator medatadata being supplied in the response

### DIFF
--- a/src/middleware/router.js
+++ b/src/middleware/router.js
@@ -221,11 +221,11 @@ function sendRequestToRoutes (ctx, routes, next) {
             if (response.headers != null && response.headers['content-type'] != null && response.headers['content-type'].indexOf('application/json+openhim') > -1) {
               // handle mediator reponse
               let payload = ''
-              response.body.on('data',function(data){
+              response.body.on('data', (data) => {
                 payload += data.toString()
               })
 
-              response.body.on('end',function(){
+              response.body.on('end', () => {
                 const responseObj = JSON.parse(payload)
                 ctx.mediatorResponse = responseObj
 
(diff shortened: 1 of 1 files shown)


src/middleware/router.js
```diff
@@ -220,15 +220,23 @@ function sendRequestToRoutes (ctx, routes, next) {
             logger.info(`executing primary route : ${route.name}`)
             if (response.headers != null && response.headers['content-type'] != null && response.headers['content-type'].indexOf('application/json+openhim') > -1) {
               // handle mediator reponse
-              const responseObj = JSON.parse(response.body)
-              ctx.mediatorResponse = responseObj
+              let payload = ''
+              response.body.on('data',function(data){
```
Zooloo2014: Looks good to me. 
Maybe just use lambda instead of function callback for consistency.

src/middleware/router.js
```diff
@@ -220,15 +220,23 @@ function sendRequestToRoutes (ctx, routes, next) {
...
+                payload += data.toString()
+              })
+
+              response.body.on('end',function(){
```
Zooloo2014: Also here, please use lambda instead of function.

src/middleware/router.js
BMartinos: Addressed in 1867f6f

src/middleware/router.js
BMartinos: Addressed in 1867f6f

PR #1046, opened 2019-07-12
OHM-829 Add back URL rewriting functionality

Zooloo2014:
URL rewriting permits the HIM to rewrite URLs that occur in the response body before passing the body back up to the invoking client.

### DIFF
--- a/src/middleware/router.js
+++ b/src/middleware/router.js
@@ -17,7 +17,7 @@ import { getGridFSBucket } from '../contentChunk'
 import { Writable, Readable } from 'stream';
 import util from 'util'
 import { brotliCompressSync } from 'zlib';
-import { makeStreamingRequest } from './streamingRouter'
+import { makeStreamingRequest, collectStream } from './streamingRouter'
 
 config.router = config.get('router')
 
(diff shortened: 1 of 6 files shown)


src/middleware/router.js
```diff
@@ -501,9 +497,19 @@ async function sendHttpRequest (ctx, route, options) {
     responseProgress: function (chunk, counter, size) {
       logger.info(`Write response CHUNK # ${counter} [ Total size ${size}]`)
     },
-    finishResponse: function () {
+    finishResponse: function (response, size) {
```
BMartinos: It doesnt seem like we make use of these parameters. Are they needed?

src/middleware/router.js
```diff
@@ -501,9 +497,19 @@ async function sendHttpRequest (ctx, route, options) {
...
       logger.info(`** END OF OUTPUT STREAM **`)
     },
+    finishResponseAsString: function(body) {
```
BMartinos: very minor: just needs a space between "function" and the bracket

test/unit/rewriteUrlsTest.js
```diff
@@ -8,7 +8,7 @@ import * as rewriteUrls from '../../src/middleware/rewriteUrls'
 import * as utils from '../../src/utils'
 
 // TODO: OHM-699 remove the x prepended to describe when url rewriting is back in support
```
BMartinos: Lets remove this comment as well. Now that the logic has been added in, this no longer applies

src/middleware/router.js
Zooloo2014: These parameters are passed in the call. They may or may not be used in the event handler. Some of the handlers have no code; the parameters are just included to document the what information is available in the call...

BMartinos: Ok cool :+1

PR #1051, opened 2019-09-09
Ohm 836 fix polling channels

bradsawadye:


src/polling.js
```diff
@@ -11,12 +11,13 @@ config.polling = config.get('polling')
...
   if (!channel.pollingSchedule) { return callback(new Error('no polling schedule set on this channel')) }
-
-  return exports.agendaGlobal.cancel({name: `polling-job-${channel._id}`}, (err) => {
-    if (err) { return callback(err) }
+  
+  try {
+    await exports.agendaGlobal.cancel({name: `polling-job-${channel._id}`})
+    
     exports.agendaGlobal.define(`polling-job-${channel._id}`, (job, done) => {
       logger.info(`Polling channel ${channel._id}`)
```
Zooloo2014: Does the the call to .define still work as a callback?

PR #1057, opened 2019-11-05
TRACE-37: Create build script for bundled release version

BMartinos:
To have a script to can reproduce the build process for the release
So that a copy can be used instantly without having to download any dependecies

TRACE-37


packaging/README.md
```diff
@@ -14,6 +14,18 @@ You must also have an environment variable set with the id of the key to use. Vi
 Login: openhie
 PPA: release
 
+# Bundled Release
+
+A bundled release will ensure all the relevant dependencies are downloaded and bundled into a built version of the OpenHIM core. Only the relevant scripts needed to run the OpenHIM core is added to the bundled release.
+
+To create a new build release execute to the below command. This does assume that your Linux distribution has the `zip` module installed
```
bradsawadye: it should be - "To create a new build release execute the below command"

packaging/build-release-zip.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
```
bradsawadye: I think we should add the command 'set -eu', so that the execution stops when there is an error. This is for cases where deleting the node modules fails and installing also fails, you will end up creating a bundle with the wrong modules

BMartinos: Done in 05b96a1

packaging/README.md
BMartinos: done in ec9ae44

packaging/README.md
```diff
@@ -14,6 +14,18 @@ You must also have an environment variable set with the id of the key to use. Vi
 Login: openhie
 PPA: release
 
+# Bundled Release
+
+A bundled release will ensure all the relevant dependencies are downloaded and bundled into a built version of the OpenHIM core. Only the relevant scripts needed to run the OpenHIM core is added to the bundled release.
+
```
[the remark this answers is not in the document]
BMartinos: done in ec9ae44

PR #1058, opened 2019-11-05
Ohm 824 fix auto retry

bradsawadye:


BMartinos:
@bradsawadye the code changes all look fine, but ive noticed something strange happening with the auto retry mechanism.

If we start with a channel without auto retry enabled and send a failed transaction, it does do a retry attmept (which is correct). If i then go an enable the channel to auto retry after the initial failed transaction i would expect any new transactions to be retried. This is however not the case as all past failed transactions get retried as well. Im not sure this is how it should be. This would mean that if we had transactions running for a year and we enable the auto retry mechanism, the past years worth of failures will be attempted to get rerun

bradsawadye:
Hey @BMartinos, I will have a look

src/middleware/streamingRouter.js
```diff
@@ -266,3 +266,29 @@ export function collectStream (readableStream) {
       })
   })
 }
+
+export function storeBodyInGridFs (body) {
```
BMartinos: This function seems to be duplicated and has been implemented at other places. Lets see if we cant re-use the function for storing data into GridFS instead of creating another function doing the same thing.

PR #1060, opened 2019-11-15
fix the handling of openhim mediator responses

bradsawadye:
Responses from openhim medaitors have to be handled diffrently. The response body contains orchestrations etc, which have to be extracted from the body. This is done for the primary route and was not being done for the secondary routes, resulting in errors

OHM-906

### DIFF
--- a/src/middleware/router.js
+++ b/src/middleware/router.js
@@ -4,7 +4,6 @@ import net from 'net'
 import tls from 'tls'
 import logger from 'winston'
 import cookie from 'cookie'
-import { Readable } from 'stream'
 import { config } from '../config'
 import * as utils from '../utils'
 import * as messageStore from '../middleware/messageStore'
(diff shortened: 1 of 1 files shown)


src/middleware/router.js
```diff
@@ -4,6 +4,7 @@ import net from 'net'
 import tls from 'tls'
 import logger from 'winston'
 import cookie from 'cookie'
+import { Readable } from 'stream'
```
BMartinos: Im not sure this method is being used anywhere and can be removed

src/middleware/router.js
```diff
@@ -329,22 +330,27 @@ const buildNonPrimarySendRequestPromise = (ctx, route, options, path) =>
...
-          if (responseObj.error) { routeObj.error = responseObj.error }
-        })
-        routeObj.response = responseObj.response
+        }
+        if (responseObj.metrics) {
+          routeObj.metrics = responseObj.metrics
+        }
+        if (responseObj.error) {
+          routeObj.error = responseObj.error
+        }
```
BMartinos: Lets maybe update these multiple if statements into a bit more cleaner way by using a ternary operator
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator
e,g `routeObj.mediatorURN = responseObj['x-mediator-urn'] ? responseObj['x-mediator-urn'] : undefined`

If it doesnt exist, it just sets the property to undefined which it would be if its not set and queried

bradsawadye: fixed in https://github.com/jembi/openhim-core-js/pull/1060/commits/80469141b8a989a1a554d41119ce1c757b84cb3d

src/middleware/router.js
bradsawadye: fixed in https://github.com/jembi/openhim-core-js/pull/1060/commits/80469141b8a989a1a554d41119ce1c757b84cb3d

PR #1063, opened 2019-11-20
Ohm 907 fix tcp route

bradsawadye:


src/middleware/retrieveTCPTransaction.js
```diff
@@ -7,6 +7,19 @@ export async function koaMiddleware (ctx, next) {
 
   ctx.body = transaction.data
   ctx.authorisedChannel = transaction.channel
+  ctx.isTcpChannel = true
```
BMartinos: @bradsawadye im not sure why this property is added to the context? was this missing?

src/middleware/streamingReceiver.js
```diff
@@ -310,20 +329,37 @@ export async function koaMiddleware (ctx, next) {
...
+    if (ctx.tcpChannelHasHttpRoute) {
+      if (collectBody) {
+        try {
+          await collectingReceiver(ctx, statusEvents)
+        } catch(err) {
+          logger.error(`collectingReceiver error: ${err}`)
+        }
+      } else {
+        streamingReceiver(ctx, statusEvents)
+      }
```
BMartinos: This functionality seems to be duplicated below, wondering if we cant move this into a re-usable function instead

test/integration/transactionsAPITests.js
```diff
@@ -283,7 +283,7 @@ describe('API Integration Tests', () => {
           .expect(201)
 
         const events = await EventModelAPI.find({})
-        events.length.should.be.exactly(6)
+        events.length.should.be.exactly(8)
```
BMartinos: what changed in the events for the count to go up by 2?

src/middleware/retrieveTCPTransaction.js
bradsawadye: @BMartinos it was not missing. This is a property that is used in the streamingReceiver middleware. The streaming middleware is purely for http requests. There are cases when we are routing from TCP to Http, and in these cases, we use the streamingReceiver middleware. This property together with 'tcpChannelHasHttpRoute' property are used to determine whether to bypass the streamingReciever or not

src/middleware/router.js
```diff
@@ -321,31 +316,35 @@ function sendRequestToRoutes (ctx, routes, next) {
 
     Promise.all(promises).then(() => {
       logger.info(`All routes completed for transaction: ${ctx.transactionId}`)
-      ctx.state.requestPromise.then(() => {
-        ctx.state.responsePromise.then(() => {
+      messageStore.initiateResponse(ctx, () => {
+        messageStore.completeResponse(ctx, () => {}).then(() => {
```
Zooloo2014: The original code here was meant to ensure that the initiateResponse and competeResponse methods in MessageStore.js only run after the initiateRequest method, which is invoked in another module (streamingRouter.js). 
This is important because the document is created by initiateRequest, and updated by completeRequest, initiateResponse and completeResponse (and so must already exist when these methods execute).
I'm not seeing how that condition is fulfilled here. Is it not necessary anymore?

src/middleware/router.js
```diff
@@ -321,31 +316,35 @@ function sendRequestToRoutes (ctx, routes, next) {
...
+          }
+          logger.debug(`Saving route events succeeded for transaction: ${ctx.transactionId}`)
+        })
+      }
     }).catch(err => {
       logger.error(err)
-      ctx.state.requestPromise.then(() => {
-        ctx.state.responsePromise.then(() => {
+      messageStore.initiateResponse(ctx, () => {
+        messageStore.completeResponse(ctx, () => {}).then(() => {
```
Zooloo2014: Same as line 319-320

src/middleware/streamingReceiver.js
bradsawadye: fixed in https://github.com/jembi/openhim-core-js/pull/1063/commits/b4fe0a7d4cb4913c1c81822ec8d5b88ad4c313a3

test/integration/transactionsAPITests.js
bradsawadye: I added an orchestration to the transaction that is used in the test and that changes the number of events recorded to 8.

src/middleware/streamingReceiver.js
```diff
@@ -319,14 +329,28 @@ export async function koaMiddleware (ctx, next) {
       ['POST', 'PUT', 'PATCH'].includes(ctx.req.method)
   }
 
-  if (collectBody && ['POST', 'PUT', 'PATCH'].includes(ctx.req.method)) {
-    try {
-      await collectingReceiver(ctx, statusEvents)
-    } catch(err) {
-      logger.error(`collectingReceiver error: ${err}`)
+  if (ctx.isTcpChannel) {
+    if (ctx.tcpChannelHasHttpRoute) {
```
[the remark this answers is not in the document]
bradsawadye: So every request that comes in, be it an http or tcp request, passes through the streamingReceiver middleware. The streamingReciever deals with http requests only. Tcp requests will only use the steamingReceiver if we are routing from tcp to http, in which case the streamingReceiver has to be used.

src/middleware/router.js
bradsawadye: Oh yeah, that functionality is not necessary anymore. It resulted in some methods being called so many times, we would end up updating the same transaction with the same info so many times

src/middleware/router.js
bradsawadye: same as above

PR #1068, opened 2020-01-02
fix the setting of the transaction status

bradsawadye:
The status 'completed with errors' is set when any secondary route fails with a statuscode of 500. This was not happening when the primary route is to an openhim mediator server, in which case the status is extracted from the response body. This commit fixes this issue


MattyJ007:
@bradsawadye Is this PR still needed? The tests are all failing

bradsawadye:
outdated pr

PR #1078, opened 2020-05-19
TRACE-125 Authentication Config Optons

MattyJ007:
Npm audit fix

TRACE-125


config/config.md
```diff
@@ -99,7 +99,19 @@ The following config option are provided by the OpenHIM. All of these options ha
   "authentication": {
     // What kind of authentication methods to allow
     "enableMutualTLSAuthentication": true,
-    "enableBasicAuthentication": true
+    "enableBasicAuthentication": true,
+    "enableCustomTokenAuthentication": false,
+    "enableJWTAuthentication": false,
+    // The secret or public key used by the encryption algorithm in signing the token
+    "jwtSecretOrPublicKey": "",
```
BMartinos, line 106: I think this should actually be `...PrivateKey`. If using a public key for this it would be possible to sign your own valid tokens if you get access to the public key
https://www.npmjs.com/package/jsonwebtoken#jwtsignpayload-secretorprivatekey-options-callback

MattyJ007: lols, changed

PR #1079, opened 2020-05-21
TRACE-126 jwt middleware

MattyJ007:


resources/certs/jwt/README.md
```diff
@@ -0,0 +1,97 @@
...
+          "secretOrPublicKey": "-----BEGIN PUBLIC KEY-----\nMIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEA4ZnfOESxGfb1MVD2coNy\n0G0bGarnKEz721MP30iyo6+YO3qzbETI8giIWGBtXD2VO49xk2miVIvZ3tAfPRnE\nsqJsOErfZ3ld5GrnLUSbUOr88cd+TTx4EqdU2dYAoc0iVEgA5UZJDLrWHM3VcHQl\nFc2F/JN78JBBZor2gWiABEFFShMN1PYmsx4IJUuE72gDVqblOLCfr+V1rT0C7iA1\n7V8lsm3jlRyBBNxdwqLvVXVcIip5/W5gQ/Ujq4KdXcC4LFR2J8idLEn4LPNsx6tA\ndHmAaBEHO9kyYgHijK+zi0b7qTYaPdrbM6siMFBh7HW6bobRqrFy5wR3zZuhg2Do\no8djtoJBHXNohxNm1D7iiNjH9jHSt9G2O5lAuDo19qb0jxMBR/ekQ3GZNTF3+C7z\nF6BXDPXY3S2Q7btYznQ6oTn/raqVaw4RiXDaBSotmOZHId2OnI5eNN4QTXr7RbOX\nzSqXf4OhiaW7Shjg0bbz6BUAKiMW6e0R3Z+JL8ZS47MxG6ibfimbZz9a0Hc5ItN2\nZK0mPPr2aEeLi/Wyaf7QB1N5IEZVj3YXJo/h5F37RnxV9IbRA201lKAVw5RbWo3Q\nur5jumzuID68U+i4rvB2JILlSykIrcGa7ffoMtTKJzTiHrKElBAEdgv4I1pUl0Js\ne2rYU8Kno94WC+34WGoUs7sCAwEAAQ==\n-----END PUBLIC KEY-----",
+          ...
+        }
+      }
+    }
+    ```
+
+    > Note: Use newline characters in the string to ensure the key structure.
+
+
```
BMartinos, line 50: We can remove this extra line here

resources/certs/jwt/README.md
```diff
@@ -0,0 +1,97 @@
...
+Second, a file containing the public key in the `pem` format can be placed in the this directory. The file name can then be specified to the config.
+
+The config can be supplied to the OpenHIM in the following ways:
+
+- Config File
+
+  - Public Key Filename.
+
+    ```json
```
BMartinos, line 16: lets change this to be `js` instead because the ellipses arent valid JSON and being highlighted in red

resources/certs/jwt/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+  - Public Key Literal.
+
+    ```sh
+    authentication_jwt_secretOrPublicKey='-----BEGIN PUBLIC KEY-----\nMIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEA4ZnfOESxGfb1MVD2coNy\n0G0bGarnKEz721MP30iyo6+YO3qzbETI8giIWGBtXD2VO49xk2miVIvZ3tAfPRnE\nsqJsOErfZ3ld5GrnLUSbUOr88cd+TTx4EqdU2dYAoc0iVEgA5UZJDLrWHM3VcHQl\nFc2F/JN78JBBZor2gWiABEFFShMN1PYmsx4IJUuE72gDVqblOLCfr+V1rT0C7iA1\n7V8lsm3jlRyBBNxdwqLvVXVcIip5/W5gQ/Ujq4KdXcC4LFR2J8idLEn4LPNsx6tA\ndHmAaBEHO9kyYgHijK+zi0b7qTYaPdrbM6siMFBh7HW6bobRqrFy5wR3zZuhg2Do\no8djtoJBHXNohxNm1D7iiNjH9jHSt9G2O5lAuDo19qb0jxMBR/ekQ3GZNTF3+C7z\nF6BXDPXY3S2Q7btYznQ6oTn/raqVaw4RiXDaBSotmOZHId2OnI5eNN4QTXr7RbOX\nzSqXf4OhiaW7Shjg0bbz6BUAKiMW6e0R3Z+JL8ZS47MxG6ibfimbZz9a0Hc5ItN2\nZK0mPPr2aEeLi/Wyaf7QB1N5IEZVj3YXJo/h5F37RnxV9IbRA201lKAVw5RbWo3Q\nur5jumzuID68U+i4rvB2JILlSykIrcGa7ffoMtTKJzTiHrKElBAEdgv4I1pUl0Js\ne2rYU8Kno94WC+34WGoUs7sCAwEAAQ==\n-----END PUBLIC KEY-----'
+    ```
+
+    > Note: Use newline characters in the string to ensure the key structure.
+
+The Public Key file must be placed in this directory. i.e `openhim-core-js/resources/certs/jwt`
```
BMartinos, line 67: Maybe a suggestion, but to move this line higher up when making mention to using a file instead of a string literal. It might get lots down here

test/unit/jwtAuthenicationTest.js
```diff
@@ -0,0 +1,663 @@
+'use strict'
+
+/* eslint-env mocha */
+
+import fs from 'fs'
+import logger from 'winston'
+import should from 'should'
+import sinon from 'sinon'
+
+import * as jwtAuthentication from '../../src/middleware/jwtAuthentication'
```
BMartinos, line 10: We can move this import down lower to keep the alphabetical ordering

test/unit/jwtAuthenicationTest.js
```diff
@@ -0,0 +1,663 @@
...
+      should(ctx.authenticated).be.null()
+      loggerStub.callCount.should.eql(1)
+    })
+
+    it('should fail and log warning on missing auth header and proceed to next Authentication middleware', async () => {
+      const ctx = {
+        authenticated: null,
+        request: {
+          header: {
+            authorization: null,
```
BMartinos, line 295: lets maybe remove the `authorization` header completely as if it wasnt supplied

test/unit/jwtAuthenicationTest.js
```diff
@@ -0,0 +1,663 @@
...
+
+    it('should fail due to missing subject field in JWT', async () => {
+      const ctx = {
+        authenticated: null,
+        request: {
+          header: {
+            authorization:
+              'Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.e30.P4Lqll22jQQJ1eMJikvNg5HKG-cKB0hUZA9BZFIG7Jk',
+          },
+        },
```
BMartinos, line 349: we can remove the trailing commas when no other properties exist

test/unit/jwtAuthenicationTest.js
```diff
@@ -0,0 +1,663 @@
...
+      }
+      const next = sandbox.spy()
+      const loggerStub = sandbox.stub(logger, 'error')
+
+      const mockJwtConfig = {
+        algorithms: 'HS256',
+        secretOrPublicKey: 'test',
```
BMartinos, line 356: we can remove the trailing commas when no other properties exist

test/unit/jwtAuthenicationTest.js
```diff
@@ -0,0 +1,663 @@
...
+        authenticated: null,
+        request: {
+          header: {
+            authorization:
+              'Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.e30.P4Lqll22jQQJ1eMJikvNg5HKG-cKB0hUZA9BZFIG7Jk',
+          },
+        },
+      }
+      const next = sandbox.spy()
+      const loggerStub = sandbox.stub(logger, 'error')
```
BMartinos, line 384: wondering if this shouldnt rather be a warning instead of an error. It could just mean that the secret is infact correct but the token used the wrong secret. This would make the request invalid as its not correctly authenticated

test/unit/jwtAuthenicationTest.js
```diff
@@ -0,0 +1,663 @@
...
+        authenticated: null,
+        request: {
+          header: {
+            authorization:
+              'Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ0ZXN0IiwiYXVkIjpbInRlc3RBdWRpZW5jZSJdLCJpc3MiOiJ0ZXN0SXNzdWVyIn0._bUjvzaXWkyYlxV81lVU1dsbZpH_jlW7sda7zsnORwg',
+          },
+        },
+      }
+      const next = sandbox.spy()
+      const loggerStub = sandbox.stub(logger, 'error')
```
BMartinos, line 416: same as above comment regarding making this log a warning instead.

test/unit/jwtAuthenicationTest.js
```diff
@@ -0,0 +1,663 @@
...
```
BMartinos, line 450: same as above comment regarding making this log a warning instead.

src/middleware/jwtAuthentication.js
```diff
@@ -0,0 +1,113 @@
+'use strict'
+
+import jwt from 'jsonwebtoken'
+import logger from 'winston'
+
+import * as client from '../model/clients'
+import * as configIndex from '../config'
+import fs from 'fs'
+import path from 'path'
```
BMartinos, line 9: we should move these two modules higher up with the other imported modules

src/middleware/jwtAuthentication.js
```diff
@@ -0,0 +1,113 @@
...
+async function authenticateToken(ctx) {
+  if (ctx.authenticated != null) {
+    return
+  }
+
+  const authHeader = ctx.request.header.authorization || ''
+  const token = TOKEN_PATTERN.exec(authHeader)
+
+  if (!token) {
+    logger.warn(`Missing or invalid 'Authorization' header`)
```
BMartinos, line 81: Lets maybe specify which auth mechanism it is that failed incase all the auth mechanisms are enabled
`Missing or invalid 'JWT Authorization' header`

src/middleware/jwtAuthentication.js
```diff
@@ -0,0 +1,113 @@
...
+
+  try {
+    const decodedToken = jwt.verify(
+      token[1],
+      resolveJwtSecretOrPublicKey(),
+      getJwtOptions()
+    )
+
+    if (!decodedToken.sub) {
+      logger.error('Invalid JWT Payload')
```
BMartinos, line 93: Maybe we need to highlight a bit more why this error happened. We can state `No JWT sub supplied in payload`

src/middleware/jwtAuthentication.js
```diff
@@ -0,0 +1,113 @@
...
+      `Could not read public key file to verify asymmetric JWT: ${error}`
+    )
+  }
+}
+
+function getJwtOptions() {
+  const jwtConfig = configIndex.config.get('authentication:jwt')
+  const jwtOptions = {}
+
+  // Algorithms can be input as an environment variable therefore the string needs to be split
```
BMartinos, line 54: this comments needs to move down to where the split is being performed

src/middleware/jwtAuthentication.js
```diff
@@ -0,0 +1,113 @@
...
+  if (jwtConfig.algorithms) {
+    jwtOptions.algorithms = jwtConfig.algorithms.split(' ')
+  } else {
+    // The jsonwebtoken package does not require this field, but allowing any algorithm to be used opens a security risk
+    throw new Error('JWT Algorithm not specified')
+  }
+
+  // Audience can be input as an environment variable therefore the string needs to be split
```
BMartinos, line 62: this comments needs to move down to where the split is being performed

src/middleware/jwtAuthentication.js
```diff
@@ -0,0 +1,113 @@
...
+      'resources',
+      'certs',
+      'jwt',
+      secretOrPublicKey
+    )
+
+    // Check file exists
+    if (fs.existsSync(publicKeyFilePath)) {
+      secretOrPublicKey = fs.readFileSync(publicKeyFilePath).toString()
+    }
```
BMartinos, line 41: Could we not just make use of the `readFileSync` method to get the file and if nothing comes back (or error thrown) the catch will handle that. Currently we are interacting with the File System twice on the exists, and then on the read

src/api/users.js
```diff
@@ -1,15 +1,16 @@
...
+import * as auditing from '../auditing'
 import * as authorisation from './authorisation'
 import * as contact from '../contact'
-import { config } from '../config'
 import * as utils from '../utils'
-import * as auditing from '../auditing'
-import crypto from 'crypto'
-import { promisify } from 'util'
+import { UserModelAPI } from '../model/users'
+import { config } from '../config'
```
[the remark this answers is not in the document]
MattyJ007: Capitalized characters come before lowercase in UTF8

http://www.fileformat.info/info/charset/UTF-8/list.htm

src/alerts.js
```diff
@@ -2,14 +2,14 @@ import logger from 'winston'
...
+import * as Channels from './model/channels'
 import * as contact from './contact'
-import { config } from './config'
-import { EventModel } from './model/events'
-import { ContactGroupModel } from './model/contactGroups'
+import * as utils from './utils'
 import { AlertModel } from './model/alerts'
+import { ContactGroupModel } from './model/contactGroups'
+import { EventModel } from './model/events'
 import { UserModel } from './model/users'
```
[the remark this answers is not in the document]
MattyJ007: Capitalized characters come before lowercase in UTF8

http://www.fileformat.info/info/charset/UTF-8/list.htm

src/api/audits.js
```diff
@@ -1,10 +1,11 @@
-import logger from 'winston'
 import atna from 'atna-audit'
+import logger from 'winston'
 import os from 'os'
-import { AuditModel, AuditMetaModel } from '../model/audits'
+
 import * as authorisation from './authorisation'
 import * as utils from '../utils'
 import * as auditing from '../auditing'
+import { AuditMetaModel, AuditModel } from '../model/audits'
```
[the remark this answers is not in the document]
MattyJ007: Capitalized characters come before lowercase in UTF8

http://www.fileformat.info/info/charset/UTF-8/list.htm

src/api/authentication.js
```diff
@@ -1,12 +1,13 @@
...
 import logger from 'winston'
-import atna from 'atna-audit'
 import os from 'os'
-import { UserModelAPI } from '../model/users'
-import { config } from '../config'
+
 import * as auditing from '../auditing'
+import { UserModelAPI } from '../model/users'
 import { caseInsensitiveRegex } from '../utils'
+import { config } from '../config'
```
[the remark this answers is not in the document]
MattyJ007: Capitalized characters come before lowercase in UTF8

http://www.fileformat.info/info/charset/UTF-8/list.htm

src/api/channels.js
```diff
@@ -1,13 +1,14 @@
 import logger from 'winston'
 import request from 'request'
+
 import * as Channels from '../model/channels'
```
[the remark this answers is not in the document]
MattyJ007: That might break tests as deconstructed imports can't be mocked :(

https://www.exratione.com/2015/12/es6-use-of-import-property-from-module-is-not-a-great-plan/

But perhaps we can have a boyscouting ticket that deals with making the import mechanism consistent - this will probably have to remove all the restructuring to allow proper testing (even though I like the neatness of it)

src/api/channels.js
```diff
@@ -1,13 +1,14 @@
...
-import { TransactionModelAPI } from '../model/transactions'
 import * as authorisation from './authorisation'
-import * as tcpAdapter from '../tcpAdapter'
-import * as server from '../server'
 import * as polling from '../polling'
 import * as routerMiddleware from '../middleware/router'
+import * as server from '../server'
+import * as tcpAdapter from '../tcpAdapter'
 import * as utils from '../utils'
+import { TransactionModelAPI } from '../model/transactions'
```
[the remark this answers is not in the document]
MattyJ007: Capitalized characters come before lowercase in UTF8

http://www.fileformat.info/info/charset/UTF-8/list.htm

src/api/keystore.js
```diff
@@ -1,9 +1,10 @@
 import pem from 'pem'
-import { KeystoreModelAPI } from '../model/keystore'
+import { promisify } from 'util'
+
 import * as authorisation from './authorisation'
 import * as utils from '../utils'
+import { KeystoreModelAPI } from '../model/keystore'
```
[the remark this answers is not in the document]
MattyJ007: Capitalized characters come before lowercase in UTF8

http://www.fileformat.info/info/charset/UTF-8/list.htm

src/api/transactions.js
```diff
@@ -1,12 +1,13 @@
...
-import * as autoRetryUtils from '../autoRetry'
+import { promisify } from 'util'
+
 import * as authorisation from './authorisation'
+import * as autoRetryUtils from '../autoRetry'
+import * as events from '../middleware/events'
 import * as utils from '../utils'
+import { ChannelModelAPI } from '../model/channels'
+import { TransactionModelAPI } from '../model/transactions'
 import { config } from '../config'
```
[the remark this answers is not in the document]
MattyJ007: Capitalized characters come before lowercase in UTF8

http://www.fileformat.info/info/charset/UTF-8/list.htm

config/config.md
```diff
@@ -102,16 +102,19 @@ The following config option are provided by the OpenHIM. All of these options ha
     "enableBasicAuthentication": true,
     "enableCustomTokenAuthentication": false,
     "enableJWTAuthentication": false,
```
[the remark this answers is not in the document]
MattyJ007: I accidentally committed my testing changes. I meant to make the default `false` as this may cause setup issues for unsuspecting users.
Users might upgrade their existing OpenHIM instances and not realize that JWTs are suddenly enabled on their instance, but without proper configuration.
Also, even if we do enable this by default along with default config this would be a security issue as there's no way to easily update the JWT default config - as we do with the default Basic Auth credentials on the first login.

src/api/transactions.js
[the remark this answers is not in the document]
BMartinos: ok cool, makes sense :+1:

src/api/users.js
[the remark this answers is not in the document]
BMartinos: :+1:

src/api/channels.js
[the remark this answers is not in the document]
BMartinos: ok cool i think thats fine. Lets do this as part of a different task in future then :+1

config/config.md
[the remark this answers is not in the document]
BMartinos: ok cool that makes sense. Lets keep it disabled by default and allow the implementer to configure it properly if they want to use it

src/middleware/jwtAuthentication.js
MattyJ007: @BMartinos Maybe we shouldn't be reading from the file system every time at all.
Could we keep the secret in an in-memory cache? The first time the secret is needed we read it - after that, we just check the cached value

BMartinos: Cool, yeah i think thats a better idea. Then we reduce the amount of IO interactions to the file system and speeds up the authentication. Maybe this should then be executed when the server starts, and not have to wait for the first request to come through

src/middleware/jwtAuthentication.js
MattyJ007: @BMartinos Is it safe to log these details? I was being cautious with how much I logged about the auth issue. Heath appears to deliberately obscure the auth error messages.
Would an attacker ever have access to the logs but not the server instance? Probably not I guess...

BMartinos: The only way they would have access to the logs if they have access to the core server to print the logs, or a user within the console where they can view the logs. The main reason for adding in the extra bit is to notify the implementer of a missing property, but not the value of what that property should be (IE clientID). 

Maybe this is giving a bit too much information as the person generating the token should be aware of all the required details for it to be considered valid

test/unit/jwtAuthenicationTest.js
MattyJ007: Is this still valid?

BMartinos: Im not entirely sure :/ to me this doesn't feel like an error as its not an exception of any kind, but rather an invalid token. If all 3 mechanisms are enabled, we will get two errors thrown every time a request comes through

BMartinos: Lets keep it as is for now and make this change in future if it becomes an issue

test/unit/jwtSecretOrPublicKeyCacheTest.js
```diff
@@ -0,0 +1,87 @@
...
+        should(error.message).equal('Could not read JWT public key file: Boom!')
+      }
+      configStub.callCount.should.eql(1)
+      fsExistsSyncStub.callCount.should.eql(1)
+      fsReadFileSyncStub.callCount.should.eql(1)
+      should(cache.getSecretOrPublicKey()).be.null()
+    })
+
+    it('should succeed on file read ', () => {
+
```
BMartinos, line 46:
```suggestion
```

PR #1080, opened 2020-05-28
TRACE-131 Update Client

MattyJ007:


src/middleware/customTokenAuthentication.js
```diff
@@ -0,0 +1,48 @@
+'use strict'
+
+import logger from 'winston'
+
+import * as client from '../model/clients'
+
+const TOKEN_PATTERN = /^ *(?:[Cc][Uu][Ss][Tt][Oo][Mm]) +([A-Za-z0-9\-._~+/]+=*) *$/
```
bradsawadye, line 7: I think we should move this constant to the constants module

MattyJ007: Why

MattyJ007: I agree with you but still, why?

src/middleware/customTokenAuthentication.js
```diff
@@ -0,0 +1,48 @@
...
+    ctx.authenticationType = 'token'
+  } catch (error) {
+    logger.error(`Custom Token could not be verified: ${error.message}`)
+    return
+  }
+}
+
+export async function koaMiddleware(ctx, next) {
+  await authenticateToken(ctx)
+  if (ctx.authenticated != null && ctx.authenticated.clientID != null) {
```
bradsawadye, line 44: shorten this if statement. ` ctx.authenticated != null to !ctx.authenticated` same for the other condition

src/middleware/customTokenAuthentication.js
```diff
@@ -0,0 +1,48 @@
...
+  return client.ClientModel.findOne({ customTokenID }).then((client) => {
+    if (!client) {
+      throw new Error('Client does not exist')
+    }
+    return client
+  })
+}
+
+async function authenticateToken(ctx) {
+  if (ctx.authenticated != null) {
```
bradsawadye, line 19: the condition here can be shortened to `!ctx.authenticated` (for consistency)

src/middleware/customTokenAuthentication.js
bradsawadye: Code modularity. We have a module where we put constants

PR #1085, opened 2020-06-29
Updated NodeJS version to use 12

BMartinos:
OpenHIM core now builds on Nodejs 12. This fix allows the build to succeed on the correct version of NodeJS

TRACE-137


MattyJ007 (approved):
LGTM

PR #1088, opened 2020-08-27
6.0.0 alpha.1

BMartinos:


BMartinos:
This new PR replaces the older branch with the incorrect version number
https://github.com/jembi/openhim-core-js/pull/1040

BMartinos:
A potential bug has been picked up with the current implementation of the routing logic. The transaction body isnt stored if the incoming data hasnt completely processed but the transaction status has been set to failed. 

This was noted with a Payload size of 425MB being sent to the HIM, but the downstream service was not running resulting in an ECONNREFUSED error and failing the transaction. Because the data is streamed as soon as it starts receiving the payload, the downstream service responded with the failure before the request payload was processed completely. It seems we discard the payload if its mid flight and an error is returned.

This prevents the transaction from being rerun once the service comes back up again as we dont have the incoming payload stored

PR #1092, opened 2020-09-02
OHM-1046 data control request storage bug

rcrichton:
This PR also includes a large number of linting fixes. It would be best to review the single commit that fixes the actual bug to get a better understanding of how it was fixed.


MattyJ007 (approved):
LGTM :D
Good catch on the bug

PR #1093, opened 2020-09-03
OHM-719 transaction gridFS migration story

MattyJ007:


src/model/transactions.js
```diff
@@ -1,4 +1,4 @@
-import { Schema, ObjectId } from 'mongoose'
+import { Schema, ObjectId, connection } from 'mongoose'
```
de-laz, line 1: @MattyJ007 It looks like `connection` is not used anywhere. Is it needed for this module?

MattyJ007: No :( How embarrassing. I've removed it in https://github.com/jembi/openhim-core-js/pull/1093/commits/adc1b97f4a2f6f5ad7b9cfb695938e2c766acfdd

de-laz:
@MattyJ007 OHM-1054 LGTM :+1:

PR #1095, opened 2020-09-08
OHM-1035 Remove logic that added transaction bodies

MattyJ007:
Transaction bodies are stored separately and can be very large. We have
rewritten the OpenHIM to stream data instead of consolidating before
orchestrating.
We are implementing this same principle with interactions with the api
to improve response times.
If a request is made for transactions, a secondary request must be made
for the potentially large transaction bodies which will then be streamed
to the requester.

OHM-1035


de-laz:
@MattyJ007 LGTM 👍

PR #1097, opened 2020-09-09
OHM-1037 - new transaction bodies endpoint

rcrichton:
Docs PR: https://github.com/jembi/openhim-docs/pull/44

### DIFF
--- a/src/contentChunk.js
+++ b/src/contentChunk.js
@@ -190,12 +190,24 @@ export const retrieveBody = async (bodyId, range) => {
 
   const fileDetails = await getFileDetails(bodyId)
 
+  if (!fileDetails) {
+    throw new Error('Could not file specified file')
+  }
+  if (range.start && range.start >= fileDetails.length) {
+    throw new Error('Start range cannot be greater than file length')
+  }
+  if (range.end && range.end > fileDetails.length) {
+    range.end = fileDetails.length
+  }
+
   const contentEncoding = fileDetails ? (fileDetails.metadata ? fileDetails.metadata['content-encoding'] : null) : null
   const decompressionStream = getDecompressionStreamByContentEncoding(contentEncoding)
 
   const bucket = getGridFSBucket()
   const downloadStream = bucket.openDownloadStream(bodyId, range)
-  downloadStream.on('error', err => { logger.error(err) })
+  downloadStream.on('error', err => {
+    logger.error(err)
+  })
 
   // apply the decompression transformation
   return { stream: downloadStream.pipe(decompressionStream), fileDetails }
--- a/src/contentChunk.js
+++ b/src/contentChunk.js
@@ -181,6 +181,24 @@ export const retrievePayload = async fileId => {
   })
 }
 
+export const retrieveBody = async (bodyId, range) => {
+  if (!bodyId) {
+    throw new Error('bodyID not supplied')
+  }
+
+  const fileDetails = await getFileDetails(bodyId)
+
+  const contentEncoding = fileDetails ? (fileDetails.metadata ? fileDetails.metadata['content-encoding'] : null) : null
+  const decompressionStream = getDecompressionStreamByContentEncoding(contentEncoding)
+
+  const bucket = getGridFSBucket()
+  const downloadStream = bucket.openDownloadStream(bodyId, range)
+  downloadStream.on('error', err => { throw err }) // TODO what would this do?
+
+  // apply the decompression transformation
+  return { stream: downloadStream.pipe(decompressionStream), fileDetails }
+}
+
 export const addBodiesToTransactions = async (transactions) => {
   if (!transactions || !Array.isArray(transactions) || transactions.length < 1) {
     return []
--- a/src/contentChunk.js
+++ b/src/contentChunk.js
@@ -178,7 +178,7 @@ export const retrieveBody = async (bodyId, range) => {
   const fileDetails = await getFileDetails(bodyId)
 
   if (!fileDetails) {
-    throw new Error('Could not file specified file')
+    throw new Error('Could not find specified file')
   }
   if (range.start && range.start >= fileDetails.length) {
     throw new Error('Start range cannot be greater than file length')
--- a/src/contentChunk.js
+++ b/src/contentChunk.js
@@ -2,6 +2,8 @@
 import mongodb from 'mongodb'
 import zlib from 'zlib'
 import { PassThrough } from 'stream'
+import logger from 'winston'
+
 import { config, connectionDefault } from './config'
 import { obtainCharset } from './utils'
 
@@ -193,7 +195,7 @@ export const retrieveBody = async (bodyId, range) => {
 
   const bucket = getGridFSBucket()
   const downloadStream = bucket.openDownloadStream(bodyId, range)
-  downloadStream.on('error', err => { throw err }) // TODO what would this do?
+  downloadStream.on('error', err => { logger.error(err) })
 
   // apply the decompression transformation
   return { stream: downloadStream.pipe(decompressionStream), fileDetails }
--- a/src/contentChunk.js
+++ b/src/contentChunk.js
@@ -178,10 +178,14 @@ export const retrieveBody = async (bodyId, range) => {
   const fileDetails = await getFileDetails(bodyId)
 
   if (!fileDetails) {
-    throw new Error('Could not find specified file')
+    const err = new Error('Could not find specified file')
+    err.status = 404
+    throw err
   }
   if (range.start && range.start >= fileDetails.length) {
-    throw new Error('Start range cannot be greater than file length')
+    const err = new Error('Start range cannot be greater than file length')
+    err.status = 416
+    throw err
   }
   if (range.end && range.end > fileDetails.length) {
     range.end = fileDetails.length
--- a/test/integration/transactionsAPITests.js
+++ b/test/integration/transactionsAPITests.js
@@ -1137,5 +1137,91 @@ describe('API Integration Tests', () => {
           .expect(403)
       })
     })
+
+    describe('*getTransactionBodyById', () => {
+      it('should stream back a full transaction body', async () => {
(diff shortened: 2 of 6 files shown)


src/contentChunk.js
```diff
@@ -181,6 +181,24 @@ export const retrievePayload = async fileId => {
...
+  }
+
+  const fileDetails = await getFileDetails(bodyId)
+
+  const contentEncoding = fileDetails ? (fileDetails.metadata ? fileDetails.metadata['content-encoding'] : null) : null
+  const decompressionStream = getDecompressionStreamByContentEncoding(contentEncoding)
+
+  const bucket = getGridFSBucket()
+  const downloadStream = bucket.openDownloadStream(bodyId, range)
+  downloadStream.on('error', err => { throw err }) // TODO what would this do?
```
rcrichton, line 196: Note to self: figure this out.

src/contentChunk.js
```diff
@@ -181,6 +183,24 @@ export const retrievePayload = async fileId => {
...
+  }
+
+  const fileDetails = await getFileDetails(bodyId)
+
+  const contentEncoding = fileDetails ? (fileDetails.metadata ? fileDetails.metadata['content-encoding'] : null) : null
+  const decompressionStream = getDecompressionStreamByContentEncoding(contentEncoding)
+
+  const bucket = getGridFSBucket()
+  const downloadStream = bucket.openDownloadStream(bodyId, range)
+  downloadStream.on('error', err => { logger.error(err) })
```
MattyJ007, line 198: @rcrichton An error is thrown here with no details when the content byte ranges are invalid (start larger than end byte) Or when in certain combinations that make no sense - i.e start=3 end=100

MattyJ007 (changes requested):
Looking just a weird content byte bug

test/integration/transactionsAPITests.js
```diff
@@ -1076,5 +1076,158 @@ describe('API Integration Tests', () => {
...
+      it('should error on an invalid range - start equals end', async () => {
+        const tx = await new TransactionModel(Object.assign({}, transactionData)).save()
+        await request(constants.BASE_URL)
+          .get(`/transactions/${tx._id}/bodies/${tx.request.bodyId}`)
+          .set('auth-username', testUtils.rootUser.email)
+          .set('auth-ts', authDetails.authTS)
+          .set('auth-salt', authDetails.authSalt)
+          .set('auth-token', authDetails.authToken)
+          .set('range', 'bytes=0-0')
+          .expect(400, 'Start range [0] cannot be greater than or equal to end [0]')
```
BMartinos, line 1172: I think this is actually leading to a bug within the expected results. This should still be considered a valid range as its indicating to start at index **0** and finish at index **0**, which will return one byte located at index **0**.

If the range was changed to `bytes=0-1` we would be getting back 2 characters and not 1. The range is inclusive of the number being defined

test/integration/transactionsAPITests.js
```diff
@@ -1076,5 +1076,158 @@ describe('API Integration Tests', () => {
...
+      it('should error on an unsupported range', async () => {
+        const tx = await new TransactionModel(Object.assign({}, transactionData)).save()
+        await request(constants.BASE_URL)
+          .get(`/transactions/${tx._id}/bodies/${tx.request.bodyId}`)
+          .set('auth-username', testUtils.rootUser.email)
+          .set('auth-ts', authDetails.authTS)
+          .set('auth-salt', authDetails.authSalt)
+          .set('auth-token', authDetails.authToken)
+          .set('range', 'bytes=1-')
+          .expect(400, 'Only accepts single ranges with both a start and an end')
```
BMartinos, line 1148: Wondering if these response codes are correct when the range is invalid, or not yet supported. There is a specific error code that handles incorrect range requests which might be a better indication of the error
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/416

src/contentChunk.js
```diff
@@ -168,6 +170,36 @@ export const retrievePayload = async fileId => {
...
 
+export const retrieveBody = async (bodyId, range) => {
+  if (!bodyId) {
+    throw new Error('bodyID not supplied')
+  }
+
+  const fileDetails = await getFileDetails(bodyId)
+
+  if (!fileDetails) {
+    throw new Error('Could not file specified file')
```
BMartinos, line 181:
```suggestion
    throw new Error('Could not find specified file')
```

src/contentChunk.js
```diff
@@ -168,6 +170,36 @@ export const retrievePayload = async fileId => {
...
+    range.end = fileDetails.length
+  }
+
+  const contentEncoding = fileDetails ? (fileDetails.metadata ? fileDetails.metadata['content-encoding'] : null) : null
+  const decompressionStream = getDecompressionStreamByContentEncoding(contentEncoding)
+
+  const bucket = getGridFSBucket()
+  const downloadStream = bucket.openDownloadStream(bodyId, range)
+  downloadStream.on('error', err => {
+    logger.error(err)
```
BMartinos, line 196: It might be better to throw error here instead of just logging out the error. Then we can inform the user of the error when trying to retrieve the body

test/integration/transactionsAPITests.js
```diff
@@ -1076,5 +1076,158 @@ describe('API Integration Tests', () => {
...
+
+      it('should error if file cannot be found', async () => {
+        const tx = await new TransactionModel(Object.assign({}, transactionData)).save()
+        await request(constants.BASE_URL)
+          .get(`/transactions/${tx._id}/bodies/222222222222222222222222`)
+          .set('auth-username', testUtils.rootUser.email)
+          .set('auth-ts', authDetails.authTS)
+          .set('auth-salt', authDetails.authSalt)
+          .set('auth-token', authDetails.authToken)
+          .expect(400, 'Could not file specified file')
```
BMartinos, line 1195:
```suggestion
          .expect(400, 'Could not find specified file')
```

test/integration/transactionsAPITests.js
```diff
@@ -1076,5 +1076,158 @@ describe('API Integration Tests', () => {
...
+        const tx = await new TransactionModel(Object.assign({}, transactionData)).save()
+        await request(constants.BASE_URL)
+          .get(`/transactions/${tx._id}/bodies/${tx.request.bodyId}`)
+          .set('auth-username', testUtils.rootUser.email)
+          .set('auth-ts', authDetails.authTS)
+          .set('auth-salt', authDetails.authSalt)
+          .set('auth-token', authDetails.authToken)
+          .set('range', 'bytes=0-0')
+          .expect(400, 'Start range [0] cannot be greater than or equal to end [0]')
+      })
```
BMartinos, line 1185: This test seems to be a duplicate of the one above, Unless im not seeing the difference? the titles are the same

test/integration/transactionsAPITests.js
BMartinos, line 1195: Should this not rather be a 404 instead of 400. The request is technically syntactically correct, but the supplied bodyId doesnt return any results. This leads me to think that i should be getting back a 404 Not Found as there is nothing i can do to the request to make it valid and return the body for the ID im supplying

BMartinos (changes requested):
Thanks @rcrichton looking good! Functional testing all works well and the code all looks good. Just left a few comments and things that might be missing.

Ive noted in the ticket description that we arent catering for multi ranges which is fine, but does this include the "wildcard" ranges? Things like `bytes=100-` (from 100 to the end of content length), and `bytes=-100` (The last 100 from content length)

test/integration/transactionsAPITests.js
rcrichton: Hmm, you might be right. I'l check it out. It's confusing as the HTTP range spec is start inclusive, end inclusive where grid fs is start inclusive end exclusive.

test/integration/transactionsAPITests.js
rcrichton: Thanks will have a look.

src/contentChunk.js
rcrichton: I agree but the issues is because we are using a stream and the error get's thrown async this will be thrown after the response has been sent and the body is streaming. I don't think there is a way to catch it and even if we do catch it, what do we do with it. So logging seemed like the only reasonable action to take. Unless you have a better idea.

test/integration/transactionsAPITests.js
rcrichton: Ok, looks like a mistake.

test/integration/transactionsAPITests.js
rcrichton: Cool that sounds better.

src/contentChunk.js
BMartinos: Ok cool, yeah thats a good point. Im not sure if this is possible but could we not add the error message as the last chuck in the stream sent to the client and then close the stream, Or just kill the stream which will fail on the client, and make the client request it again, hopefully with no errors. Not sure if this is possible or feasible though. Just thinking in terms of notifying the client that something happened instead of only getting half the requested payload back. But maybe this can be implemented another time

rcrichton (commented):
Note to self: update the API docs

test/integration/transactionsAPITests.js
rcrichton: Good catch, resovled.

src/contentChunk.js
rcrichton: I don't think we should pollute the stream with random error text. It might break the response even more. Especially if we just append text to what the client thinks is an image for example. I couldn't think of a good way to notify the user, so I think this is the best we can do.

BMartinos: Ok cool that makes sense, Lets leave this as is for now then and can re-look into this more if this is picked up as an issue for clients. For now, The console will be the only real client making use of this endpoint so should be fine for the time being

rcrichton:
Docs PR has been created here: https://github.com/jembi/openhim-docs/pull/44

BMartinos (approved):
@rcrichton thanks, this is working well :+1:

PR #1106, opened 2020-11-11
Updated package version with a patch increase

BMartinos:
To indicate the dependecy updates
and the bug fix on the clientId/role name not allowed to be the same


MattyJ007 (approved):
:+1:

PR #1108, opened 2021-01-05
fix the channel report

bradsawadye:
The channel name was not being displayed on the report as the channel name and the channel id were not being injected into the report template. The two properties are in a nested data object, and it seems the HandleBar module which does the compiling and injection was failing to retrieve the properties. Logic has been added to put the propertes on the first layer where they can be easily retrieved. Some minor refactor has also been done

MOM-651


MattyJ007 (approved):
Nice one

PR #1115, opened 2021-03-25
fix tls authentication issue

bradsawadye:
The success message should only be logged out when tls authentication is successful.  Logging out the message was causing an error to be thrown when tls auth fails

DATO-8


de-laz (approved):
@bradsawadye LGTM :+1:

PR #1135, opened 2021-07-05
Add logic for matching the request http method

bradsawadye:
THis enables requests to be differentiated using the http method

OHM-1078

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "openhim-core",
   "description": "The OpenHIM core application that provides logging and routing of http requests",
-  "version": "7.0.0",
+  "version": "7.0.1",
   "main": "./lib/server.js",
   "bin": {
     "openhim-core": "./bin/openhim-core.js"
--- a/src/middleware/authorisation.js
+++ b/src/middleware/authorisation.js
@@ -49,6 +49,9 @@ export async function authorise (ctx, done) {
     // authorisation succeeded
     ctx.authorisedChannel = channel
     logger.info(`The request, '${ctx.request.path}' is authorised to access ${ctx.authorisedChannel.name}`)
+  } else if(!channel) {
+    // Channel not found
+    ctx.response.status = 404
   } else {
     // authorisation failed
     ctx.response.status = 401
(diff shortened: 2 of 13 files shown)


src/middleware/authorisation.js
```diff
@@ -49,6 +49,9 @@ export async function authorise (ctx, done) {
     // authorisation succeeded
     ctx.authorisedChannel = channel
     logger.info(`The request, '${ctx.request.path}' is authorised to access ${ctx.authorisedChannel.name}`)
+  } else if(!channel) {
```
MattyJ007, line 52: @rcrichton 
What are your thoughts on auth here.

Returning 401 appears to be revealing the existence of a channel to unauthenticated users? I think this is a problem.
Returning 404 in `no channel` and `auth failed` would mean "No channel exists for your credentials" - which protects the existence of channels from unauthenticated users.
Or is this overboard.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "openhim-core",
   "description": "The OpenHIM core application that provides logging and routing of http requests",
-  "version": "7.0.0",
+  "version": "7.0.1",
```
MattyJ007, line 4: I think this should be 7.1.0 - this adds new functionality to OpenHIM without affecting the previous versions. It's not really a patch as this was not a defined feature of the OpenHIM AFAIK

MattyJ007 (commented):
@bradsawadye looking good 👍🏼
One minor comment on the version - also a possible auth improvement pending Ryan
Any chance you could add an integration test? If it's not too much effort

package.json
MattyJ007: NVM - I just checked and this is definitely something we were supposed to have supported - so 7.0.1 👍🏼

src/middleware/authorisation.js
rcrichton: Hmm, they would have to have guessed the path already, so I don't think this is too big a deal. It might also make issues harder to debug for a user. My feeling is that we leave it as is.

MattyJ007 (approved):
👍🏼

PR #1141, opened 2021-09-13
OHM-1084 Update core validation for audit log patientID filter

michaelloosen:
To prevent special characters being used in the patientID field (fix a bug)
OHM-1084 OHM-927


MattyJ007 (approved):
Nice one 👍🏼

PR #1153, opened 2021-10-18
Ohm 1095

Ken473:
This is the first pass of the refactoring for the Channels and the clients. This is for you to see and comment if you're happy with the approach we have taken. We are going to add more refactorings and apply those to the rest of the entities.


michaelloosen:
Please fix linting and look at failing tests

MattyJ007:
Also, please include a message body and the ticket in your commit as well - not just a heading.
The body should explain why you are making the change. Poor commit messages make maintaining a codebase difficult - especially in a large scale API change.

PR #1157, opened 2021-11-05
Fixed IP whitelist authorisation

cozej4:
From the Documentation.

> Whitelisted IP Addresses - A list of IP addresses that will be given access without authentication required.

Fixed issues in authorizing requests from whitelisted IP which should not require additional Client Authorisation


MattyJ007:
Thanks for the contribution @cozej4 💯 
We will review this as soon as we can.

cozej4:
> Thanks for the contribution @cozej4 💯 We will review this as soon as we can.

Thanks @MattyJ007

PR #1165, opened 2022-08-24
Plat 249 reduce rerun time

bradsawadye:


MattyJ007:
Failed build = 🍰

src/api/transactions.js
```diff
@@ -193,130 +195,242 @@ export async function getTransactions(ctx) {
...
+    filters = parseTransactionFilters(filters)
+
+    // execute the query
+    let transactions = await TransactionModelAPI.find(filters, projectionFiltersObject)
+      .skip(filterSkip)
+      .limit(parseInt(filterLimit, 10))
+      .sort({'request.timestamp': -1})
+      .exec()
+
+    if (filterRepresentation == 'bulkrerun') {
```
michaelloosen, line 207: Should this be `===`?

src/tasks.js
```diff
@@ -19,6 +19,10 @@ let activeTasks = 0
 export async function findAndProcessAQueuedTask() {
   let task
   try {
+    if (activeTasks > 3) {
```
michaelloosen, line 22: Should we be hardcoding this 3 or should we rather have it configurable like the `task_transactions_length`

src/tasks.js
```diff
@@ -118,8 +122,22 @@ async function processNextTaskRound(task) {
   const transactions = Array.from(
     task.transactions.slice(nextI, nextI + task.batchSize)
   )
+  const modifiedTask = await TaskModel.findById(task._id)
+
+  if (modifiedTask.status == 'Paused') {
```
michaelloosen, line 127: Should this be `===`?

test/unit/tasksTest.js
```diff
@@ -380,6 +380,9 @@ describe('Rerun Task Tests', () => {
     it('will process empty tasks', async () => {
       const originalTask = await createTask()
       await tasks.findAndProcessAQueuedTask()
+      await new Promise(resolve => {
```
michaelloosen, line 383: Is this basically just a sleep? Shouldn't we await the actual execution somehow instead of giving a hardcoded time?

test/integration/transactionsAPITests.js
```diff
@@ -16,6 +16,7 @@ import {ChannelModel} from '../../src/model/channels'
 import {EventModelAPI} from '../../src/model/events'
 import {TransactionModel} from '../../src/model/transactions'
 import {config} from '../../src/config'
+import { TaskModel } from '../../src/model'
```
michaelloosen, line 19: Inconsistent spacing

src/api/transactions.js
```diff
@@ -193,130 +195,242 @@ export async function getTransactions(ctx) {
...
+  if (pauseQueue) {
+    taskObject.status = 'Paused'
+  }
+
+  const task = await new TaskModelAPI(taskObject).save()
+
+  logger.info(`Rerun task with id ${task._id} created!`)
+
+  if (page < pages) {
+    return createRerunTasks(filters, batchSize, email, ++page, pages, pauseQueue, filterLimit)
```
michaelloosen, line 317: Not sure if we can do this differently, but I think we might eventually run into a large enough dataset for the recursion to use too much memory, probably not an issue now, but just putting it here for future reference. Let me know what you think.

michaelloosen (changes requested):
Looks good, minor fixes

config/default.json
```diff
@@ -39,15 +39,16 @@
     "maxPayloadSizeMB": 50,
     "truncateSize": 15000,
     "truncateAppend": "\n[truncated ...]",
-    "authenticationTypes": ["token"]
+    "authenticationTypes": ["basic", "token"]
```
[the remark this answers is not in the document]
MattyJ007: I think this should be here by default since we never use tokens...
Thought perhaps it's a security risk - needs discussion

test/integration/transactionsAPITests.js
MattyJ007: 🍰

src/api/transactions.js
bradsawadye: We are unlikely to run into issues here. The recursive function does not have local variables. The variables used are global and each time the function is called it replaces the values of those variables and once the recursion is done the variables are set to null

test/unit/transactionsTest.js
```diff
@@ -47,3 +49,43 @@ describe('calculateTransactionBodiesByteLength()', () => {
...
+  it('should create rerun task', async () => {
+    await TransactionModel(transaction).save()
+    await transactions.createRerunTasks({}, 1, userEmail, 0, 0, 'Paused', 1)
+    const tasks = await TaskModel.find()
+    tasks.length.should.be.exactly(1)
+  })
+
+  it('should create multiple rerun tasks', async () => {
+    await TransactionModel(transaction).save()
+    await TransactionModel(transaction).save()
```
marrouchi, line 85: Duplicated line ?

src/tasks.js
```diff
@@ -118,8 +122,22 @@ async function processNextTaskRound(task) {
...
+    )
+    TaskModel.findByIdAndUpdate({_id: task._id}, {
+      remainingTransactions: task.remainingTransactions
+    })
+
+    return
+  }
 
   const promises = transactions.map(transaction => {
+    task.remainingTransactions--
```
marrouchi, line 139: Better to decrement once the promise is resolve ? wdyt ?

marrouchi (changes requested):
I hope the review is beneficial :pray:

test/unit/transactionsTest.js
bradsawadye: It not duplication as such. We actually want to create two transactions here

src/alerts.js
```diff
@@ -257,7 +257,7 @@ const findTransactionsMaxRetried = (channel, alert, dateFrom, callback) =>
       status: 500,
       autoRetryAttempt: channel.autoRetryMaxAttempts
     },
-    {transactionID: 'transactionID'}
+    {transactionID: 1}
```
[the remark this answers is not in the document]
bradsawadye: i think its fine leaving it like this. We only want the transaction ids here

src/tasks.js
bradsawadye: Decrementing after will result in the rerunning process being slow. This is refactoring that we did to reduce the length of rerunning when we do bulk reruns for data recovery

marrouchi (approved):

michaelloosen (approved):
LGTM

PR #1180, opened 2023-01-30
PLAT 638 Adjust login wih session and cookies

nour-borgi:
In this PR: 
- Added koa-passport 
- Added koa-session
- Added koa-compose to handle multiple middlewares 

Now authentication can be done by either "basic" or "local" types.
"local" means through the UI with username and password, this will create a session and set cookies in the browser.
"basic" means with basic auth either through browser or postman by giving also username and password.

In Mongo, added passport and session tables to store auth info.


package.json
```diff
@@ -67,6 +72,10 @@
     "mongoose-patch-history": "^2.0.0",
     "nconf": "0.10.0",
     "nodemailer": "^6.6.3",
+    "nodemon": "^2.0.20",
```
marrouchi, line 75: This should be a dev dependency + would be convinient to add a command in the scripts to watch and hot-reload changes using nodemon.

src/api/passport.js
```diff
@@ -0,0 +1,23 @@
+import bcrypt from 'bcryptjs';
+
+export const validatePassword = function (passport, password, next) {
+  bcrypt.compare(password, passport.password, next);
+};
+
+export async function hashPassword(password) {
```
marrouchi, line 7: Shouldn't this be a helper function ? we can move it to "utils" folder  ?

src/api/passport.js
```diff
@@ -0,0 +1,23 @@
...
+  var salt = 8;
+
+  if (password) {
+    return new Promise((resolve, reject) => {
+      bcrypt.hash(password, salt, function (err, hash) {
+        if (err) {
+          console.error(err);
```
marrouchi, line 14: Do we have a "koa" or app logger somewhere that we can reuse ?

src/api/passport.js
```diff
@@ -0,0 +1,23 @@
+import bcrypt from 'bcryptjs';
+
+export const validatePassword = function (passport, password, next) {
+  bcrypt.compare(password, passport.password, next);
+};
+
+export async function hashPassword(password) {
+  var salt = 8;
```
marrouchi, line 8: Salt is too short, and should be rather something we pull from the config.

src/api/passport.js
```diff
@@ -0,0 +1,23 @@
...
+      bcrypt.hash(password, salt, function (err, hash) {
+        if (err) {
+          console.error(err);
+          reject(err);
+        }
+        resolve(hash);
+      });
+    });
+  } else {
+    throw new Error("Password wasn't provided");
```
marrouchi, line 21:
```suggestion
    return Promise.reject(new Error("Password wasn't provided"));
```

src/api/authentication.js
```diff
@@ -164,15 +73,15 @@ function isAuthenticationTypeEnabled(type) {
...
+  let user = null
   // First attempt basic authentication if enabled
   if (user == null && isAuthenticationTypeEnabled('basic')) {
-    user = await authenticateBasic(ctx)
+    user = await authenticateBasic(ctx, next)
   }
   // Otherwise try token based authentication if enabled
-  if (user == null && isAuthenticationTypeEnabled('token')) {
-    user = await authenticateToken(ctx)
+  if (user == null && isAuthenticationTypeEnabled('local') && ctx.req.user) {
```
marrouchi, line 83: is ctx.req.user a reference to the req.session object ? or is there a middleware that performs "ctx.req.user = getUserbySessionId(req.session)"

src/api/users.js
```diff
@@ -269,9 +270,17 @@ export async function updateUserByToken(ctx, token) {
   }
 
   try {
-    await UserModelAPI.findOneAndUpdate({token}, userUpdateObj)
-    ctx.body = 'Successfully set new user password.'
-    return logger.info(`User updated by token ${token}`)
+    const userToBeUpdated = await UserModelAPI.findOne({token})
```
marrouchi, line 273: If this is for password reset, then we should add a restriction for the users that have "local" provider, right? Same goes for the profile update.

src/koaApi.js
```diff
@@ -25,15 +27,41 @@ import * as tasks from './api/tasks'
...
-  // Create an instance of the koa-server and add a body-parser
+  // Create an instance of the koa-server
   const app = new Koa()
-  app.use(cors({allowMethods: 'GET,HEAD,PUT,POST,DELETE'}))
+
+  // Add cors options
+  app.use(cors({allowMethods: 'GET,HEAD,PUT,POST,DELETE', credentials: true}))
+
+  // Configure Sessions Middlewares
+  app.keys = [ 'r8q,+&1LM3)CD*zAGpx1xm{NeQhc;#' ]
```
marrouchi, line 42: Is this the session salt ? Could we make this a configuration ?

src/koaApi.js
```diff
@@ -25,15 +27,41 @@ import * as tasks from './api/tasks'
...
+  app.use(
+    session({
+      maxAge: 2 * 60 * 60 * 1000, // 2 hours
+      resave: false,
+      secure: true,
+      httpOnly: false,
```
marrouchi, line 48: httpOnly: true

src/koaApi.js
```diff
@@ -25,15 +27,41 @@ import * as tasks from './api/tasks'
...
-  app.use(cors({allowMethods: 'GET,HEAD,PUT,POST,DELETE'}))
+
+  // Add cors options
+  app.use(cors({allowMethods: 'GET,HEAD,PUT,POST,DELETE', credentials: true}))
+
+  // Configure Sessions Middlewares
+  app.keys = [ 'r8q,+&1LM3)CD*zAGpx1xm{NeQhc;#' ]
+  app.use(
+    session({
+      maxAge: 2 * 60 * 60 * 1000, // 2 hours
```
marrouchi, line 45: Let's make this a configuration.

src/api/users.js
```diff
@@ -66,6 +64,12 @@ export async function authenticate(ctx, email) {
   }
 }
 
+export const logout =  async function(ctx) {
+  ctx.session = null;
```
marrouchi, line 68: Does passport automatically delete the session document from the DB ?

src/middleware/sessionStore.js
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Session Store
+ *
+ * To be able to store the session in a Mongo collection instead of just saving it in memory
+ * 
+ */
+
+const SessionSchema = new Schema({
+  _id: String,
+  data: Object,
```
marrouchi, line 13: Is there a possiblity with mongo to have a ref to the user object? Rather than having a json that contains the user id ?Or is it the way passport wants it ?

src/middleware/sessionStore.js
```diff
@@ -0,0 +1,56 @@
...
+  updatedAt: {
+    default: new Date(),
+    type: Date
+  }
+});
+
+class MongooseStore {
+  constructor() {
+    this.sessionModel = connectionDefault.model('Session', SessionSchema)
+    this.session = connectionAPI.model('Session', SessionSchema)
```
marrouchi, line 23: Why do we need two models ?

src/model/passport.js
```diff
@@ -0,0 +1,69 @@
...
+    // key, typically an ID. These two fields are used as the main means of
+    // identifying a passport and tying it to a local user.
+    //
+    // The "tokens" field is a JSON object used in the case of the OAuth stan-
+    // dards. When using OAuth 1.0, a `token` as well as a `tokenSecret` will
+    // be issued by the provider. In the case of OAuth 2.0, an `accessToken`
+    // and a `refreshToken` will be issued.
+    provider: {
+        type: String,
+        enum: ['local', 'keycloak'],
```
marrouchi, line 55:
```suggestion
        enum: ['local', 'openid'],
```

src/passport.js
```diff
@@ -0,0 +1,73 @@
...
+        )
+      } else if (key === 'basic' && strategies.basic) {
+        Strategy = strategies[key].strategy
+        passport.use(
+          new Strategy(
+            async (username, password, done) =>
+              await basic.login(username, password, done)
+          )
+        )
+      }
```
marrouchi, line 45: Else we throw an error "Unknown auth provider"

src/protocols/local.js
```diff
@@ -0,0 +1,224 @@
...
+}
+
+/**
+ * Register a new user
+ *
+ * This method creates a new user from a specified email, username and password
+ * and assign the newly created user a local Passport.
+ *
+ */
+export const createUser = async function (_user) {
```
marrouchi, line 89: Shouldn't we place this in the user model file ?

src/protocols/local.js
```diff
@@ -0,0 +1,224 @@
...
+}
+
+/**
+ * Update user
+ *
+ * This method updates an user based on its id or username if id is not present
+ * and assign the newly created user a local Passport.
+ *
+ */
+export const updateUser = async function (_user) {
```
marrouchi, line 117: Same as the previous comment

marrouchi (changes requested):
Amazing work @nour-borgi , left a couple of comments !

package.json
nour-borgi: Thanks yes it should be dev dependency 👍🏼, and I already added a command to hot-reload

src/api/passport.js
nour-borgi: yes sure :)

src/api/passport.js
nour-borgi: cool done :)

src/api/passport.js
nour-borgi: yes fixed that! thanks

src/api/passport.js
nour-borgi: cool done :)

src/api/users.js
nour-borgi: nope, I need to do that manually, I will address this

src/api/users.js
nour-borgi: I can see that a user created using the openid can have a password also, why not ?

src/koaApi.js
nour-borgi: cool done :)

src/koaApi.js
nour-borgi: cool done :)

src/middleware/sessionStore.js
nour-borgi: I don't think I have the possibility to change that
According to the code passport is accessing a key first and then user:

image


data should be : { passport : { user : ... } } with passport is the default key

src/middleware/sessionStore.js
nour-borgi: I used the same pattern as the other models.

there is different connection types in the code one with options and one without, 
so I stuck to the same way the other models were defined

src/model/passport.js
nour-borgi: yes I will update that in the next PR, but removed it now

src/protocols/local.js
nour-borgi: cool done :)

src/protocols/local.js
nour-borgi: cool

src/koaApi.js
nour-borgi: cool

src/middleware/sessionStore.js
marrouchi: So one is used for UI and the other for the API, we don't need the session for the API. right ?

src/server.js
```diff
@@ -351,19 +347,19 @@ if (cluster.isMaster && !module.parent) {
...
-    agenda.on('start', job =>
-      logger.info(
-        `starting job: ${job.attrs.name}, Last Ran at: ${job.attrs.lastRunAt}`
-      )
-    )
+    // agenda.on('start', job =>
+      // logger.info(
+      //   `starting job: ${job.attrs.name}, Last Ran at: ${job.attrs.lastRunAt}`
+      // )
+    // )
```
marrouchi, line 354: Let's remove this ?

src/server.js
```diff
@@ -351,19 +347,19 @@ if (cluster.isMaster && !module.parent) {
...
     agenda.on('fail', (err, job) =>
       logger.error(`Job ${job.attrs.name} failed with ${err.message}`)
     )
 
-    agenda.on('complete', job =>
-      logger.info(`Job ${job.attrs.name} has completed`)
-    )
+    // agenda.on('complete', job =>
+      // logger.info(`Job ${job.attrs.name} has completed`)
+    // )
```
marrouchi, line 362: This as well ? or is it needed ?

src/middleware/sessionStore.js
nour-borgi: cool done

src/server.js
nour-borgi: cool done

src/server.js
nour-borgi: coo done

src/middleware/tlsAuthentication.js
```diff
@@ -2,7 +2,7 @@
 
 import logger from 'winston'
 import pem from 'pem'
-import {rootCas as rootCAs} from 'ssl-root-cas/latest'
+import {rootCas as rootCAs} from 'ssl-root-cas'
```
marrouchi, line 5: Does this need updating ?

src/model/passport.js
```diff
@@ -0,0 +1,113 @@
...
+    // Associations
+    //
+    // Associate every passport with one, and only one, user.
+    user: { type: Schema.Types.ObjectId, ref: 'User' },
+})
+
+
+// compile the Passport Schema into a Model
+export const PassportModelAPI = connectionAPI.model('Passport', PassportSchema)
+export const PassportModel = connectionDefault.model('Passport', PassportSchema)
```
marrouchi, line 69: Do we really need 2 models here ?

marrouchi (approved):

src/model/passport.js
nour-borgi: done :)

src/middleware/tlsAuthentication.js
nour-borgi: I was getting error about not finding latest

rcrichton: We need to use latest here for security reasons. What error are you getting?

https://www.npmjs.com/package/ssl-root-cas#using-the-latest-certificates

rcrichton (changes requested):
Looks very good to me. A few comments.

src/api/authentication.js
```diff
@@ -36,108 +34,20 @@ const auditingExemptPaths = [
...
-    ctx.throw(
-      401,
-      `Password did not match expected value, denying access to API, the request was made by ${email} from ${ctx.request.host}`,
-      {email}
-    )
-  }
-  return user
-}
-
-async function authenticateToken(ctx) {
```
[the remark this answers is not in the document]
rcrichton: Note, the console doesn't have to support both, just the core. So no changes are necessary on the console.

src/middleware/tlsAuthentication.js
nour-borgi: image

I will bring it back anyway, ad it is saying ('not a big deal')

src/api/authentication.js
```diff
@@ -36,108 +34,20 @@ const auditingExemptPaths = [
...
-    // not authenticated - token mismatch
-    ctx.throw(
-      401,
-      `API token did not match expected value, denying access to API, the request was made by ${email} from ${ctx.request.host}`,
-      {email}
-    )
-  }
+async function authenticateBasic(ctx, next) {
+  // Basic auth using middleware
+  await passport.authenticate('basic', function (err, user) {
```
bradsawadye, line 39: The function here takes in a callback. It does return a promise we can `await`

src/api/authentication.js
```diff
@@ -36,108 +34,20 @@ const auditingExemptPaths = [
...
+    if (err) {
+      return logger.info(err.message)
+    } else if (user) {
+      ctx.req.user = user
+      ctx.body = 'User Authenticated Successfully'
+      ctx.status = 200
+      return ctx.req.user
```
bradsawadye, line 46: i dont think the returns we have here are necessary. The function takes in the context (ctx) and adds the `req.user` to it. If want to access the users details you can just check the ctx.

src/api/authentication.js
bradsawadye: next is also not being used here

src/api/authentication.js
```diff
@@ -36,108 +34,20 @@ const auditingExemptPaths = [
...
+    }
+  })(ctx, next)
 
-  return user
+  return ctx.req.user || null
```
bradsawadye, line 50: Do we have to return null when the ctx.req.user is undefined (or falsy)? Cant we just return undefined(or false)?

config/default.json
```diff
@@ -39,7 +39,7 @@
     "maxPayloadSizeMB": 50,
     "truncateSize": 15000,
     "truncateAppend": "\n[truncated ...]",
-    "authenticationTypes": ["basic", "token"]
```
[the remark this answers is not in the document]
nour-borgi: cool done

src/model/passport.js
nour-borgi: I m following the other models for consistency and connectionAPI contains other config than the connectionDefault
so I was sticking to the same way the other models were done

src/api/authentication.js
nour-borgi: yes, the condition afterwards is if (user == null)

src/api/authentication.js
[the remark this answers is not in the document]
rcrichton: @nour-borgi we spoke about this in the architecture forum. The result of that is that we feel we need to take the safe approach which is to support the token based approach for one more version while we implement these passport changes but mark it as deprecated, then remove it in a subsequent version.

So this mean we need the following changes on this PR:
* Implement a passport auth strategy that implements the old token approach
* When this method is used add a log messages that states it is being deprecated
* Implement a migration to move existing users to the new db structure - this can be done [here as we have support for migrations](https://github.com/jembi/openhim-core-js/blob/master/src/upgradeDB.js) (this needs to be done regardless of if we keep the token auth as we need to give people a way to upgrade their db)

cc @marrouchi @michaelloosen

package.json
```diff
@@ -67,6 +71,9 @@
     "mongoose-patch-history": "^2.0.0",
     "nconf": "0.10.0",
     "nodemailer": "^6.6.3",
+    "passport": "^0.5.0",
```
bradsawadye, line 74: This can be removed since we are now using koa-passport

src/api/authentication.js
nour-borgi: I will update this ..

src/api/authentication.js
nour-borgi: well noted, will be updated

src/api/authentication.js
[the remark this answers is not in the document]
nour-borgi: Well noted. Thanks

src/koaApi.js
```diff
@@ -39,11 +39,11 @@ export function setupApp(done) {
   app.use(cors({allowMethods: 'GET,HEAD,PUT,POST,DELETE', credentials: true}))
 
   // Configure Sessions Middleware
-  app.keys = [config.authentication.sessionKey]
+  app.keys = [config.api.sessionKey || 'r8q,+&1LM3)CD*zAGpx1xm{NeQhc;#']
```
rcrichton, line 42: I think defaulting this might be dangerous. As we also default it in the config. If for some reason the config isn't being read then this exposed key will be used. Can we remove this default value.

rcrichton (changes requested):
Looks good on a quick scan, just one comment. I'l try go through and test it and give more feedback when I am able.

src/koaApi.js
nour-borgi: done:)

src/api/authentication.js
```diff
@@ -160,20 +76,26 @@ function getEnabledAuthenticationTypesFromConfig(config) {
...
+  let user = null
+
+  // First attempt local authentication if enabled
+  if (user == null && ctx.req.user) {
+    user = ctx.req.user
   }
-  // Otherwise try token based authentication if enabled
-  if (user == null && isAuthenticationTypeEnabled('token')) {
+  // Otherwise try token based authentication if enabled (@deprecated)
+  if (user == null) {
```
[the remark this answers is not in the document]
nour-borgi: I changed the logic to be handled in the same way for all the authentication types (in passport.js)

rcrichton (approved):
LGTM, I've tested this and all appears to work well. Just added a tiny comment.

src/api/authentication.js
```diff
@@ -195,9 +117,50 @@ function handleAuditResponse(err) {
 }
 
 export async function authenticate(ctx, next) {
-  let user
   try {
-    user = await authenticateRequest(ctx)
+    // Authenticate Request either by basic or local
```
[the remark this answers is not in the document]
nour-borgi: cool done

PR #1188, opened 2023-04-06
PLAT-707 Add kafka integration

nour-borgi:


src/middleware/kafkaProducer.js
```diff
@@ -0,0 +1,81 @@
...
+    const {message, ...extra} = log
+    logger[toWinstonLogLevel(level)]({
+      message,
+      extra
+    })
+  }
+}
+
+export class KafkaProducer {
+  static details = {}
```
marrouchi, line 29:
```suggestion
  _options = {}
```
This shouldn't be static, otherwise it will hold a single value across all producers.
Renaming to "options" is better ;)

src/middleware/kafkaProducer.js
```diff
@@ -0,0 +1,81 @@
...
+      const kafka = new Kafka({
+        brokers: kafkaBrokers,
+        clientId: clientId,
+        requestTimeout: timeout,
+        connectionTimeout: timeout,
+        logLevel: logLevel.DEBUG,
+        logCreator: kafkaLogger
+      })
+
+      KafkaProducer.details = {
```
marrouchi, line 46:
```suggestion
      this._options = {
```

src/middleware/kafkaProducer.js
```diff
@@ -0,0 +1,81 @@
...
+
+      this._producer = kafka.producer()
+
+      this._producer.on(this._producer.events.DISCONNECT, () => {
+        this._isConnected = false
+      })
+    }
+  }
+
+  get getDetails() {
```
marrouchi, line 60:
```suggestion
  get options() {
```
you could do kafkaProducer.options directly.

src/middleware/kafkaProducer.js
```diff
@@ -0,0 +1,81 @@
...
+    return KafkaProducer.details
```
marrouchi, line 61:
```suggestion
    return this._options
```

src/middleware/kafkaProducer.js
```diff
@@ -0,0 +1,81 @@
...
+        clientId: clientId,
+        requestTimeout: timeout,
+        connectionTimeout: timeout,
+        logLevel: logLevel.DEBUG,
+        logCreator: kafkaLogger
+      })
+
+      KafkaProducer.details = {
+        kafkaBrokers: brokers,
+        kafkaClientId: clientId,
```
marrouchi, line 48:
```suggestion
        brokers: brokers,
        clientId: clientId,
```
Since it's within the KafkaProducer class, it's okay not to have a prefix. We know it's about kafka :)

src/middleware/kafkaProducer.js
```diff
@@ -0,0 +1,81 @@
...
+    return KafkaProducer.details
+  }
+
+  get isConnected() {
+    return this._isConnected
+  }
+
+  async connect() {
+    await this._producer.connect()
+    this._isConnected = true
```
marrouchi, line 70:
```suggestion
    if (!this._isConnected) {
      try {
        await this._producer.connect()
        this._isConnected = true
       } catch() {
          // logging
       }
    }
    return this._isConnected;
```

src/middleware/kafkaProducer.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  async disconnect() {
+    await this._producer.disconnect()
```
marrouchi, line 74: We may need to try/catch here and also test if we are connected (opposite condition as for connect() )

src/middleware/kafkaProducerSet.js
```diff
@@ -0,0 +1,57 @@
+import logger from 'winston'
+
+import {KafkaProducer} from './kafkaProducer.js'
+
+export class KafkaProducerSet {
+  static kafkaSet = []
```
marrouchi, line 6: Let's change this to "{}" instead of array, to allow access by key (string computed from clientId + broker). This prevent us from doing loops to search and thus enhance performance.

src/middleware/kafkaProducerSet.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  static async getProducer(details) {
+    const kafkaExist = this.findKafkaInstance(details)
+
+    if (kafkaExist) {
+      if (!kafkaExist.isConnected) {
+        await kafkaExist.connect()
+      }
+      return kafkaExist.producer
```
marrouchi, line 25:
```suggestion
      if (!kafkaExist.isConnected) {
         throw new Error('Kafka Producer failed to connect.')
      }
      return kafkaExist
```

marrouchi (changes requested):
Left a couple of comments. Looks good overall

We may rename KafkaProducerSet to KafkaProducerManager.

src/middleware/kafkaProducer.js
nour-borgi: yes 😄 👍🏼 done

src/middleware/kafkaProducer.js
nour-borgi: done 👍🏼

src/middleware/kafkaProducer.js
nour-borgi: done 👍🏼

src/middleware/kafkaProducer.js
nour-borgi: done 👍🏼

src/middleware/kafkaProducer.js
nour-borgi: okay cool :)

src/middleware/kafkaProducerSet.js
nour-borgi: cool done 👍🏼

src/middleware/kafkaProducerSet.js
nour-borgi: yes thanks done

src/middleware/kafkaProducer.js
nour-borgi: I didn't add a catch here to save the original error message

src/middleware/kafkaProducer.js
nour-borgi: done :)

src/middleware/KafkaProducerManager.js
rcrichton, line 1: This file isn't technically a middleware, I'd suggest moving to the root of the `src` folder or another folder that makes more sense.

src/middleware/kafkaProducer.js
rcrichton, line 1: This file isn't technically a middleware, I'd suggest moving to the root of the src folder or another folder that makes more sense.

src/model/channels.js
```diff
@@ -28,21 +43,18 @@ const RouteDef = {
...
   forwardAuthHeader: {
     type: Boolean,
     default: false
-  }
+  },
+  waitPrimaryResponse: Boolean,
+  statusCodesCheck: String,
+  kafkaClientId: String,
+  // Kafka route definition
+  kafkaBrokers: {
```
rcrichton, line 54: I think the broker should be a global configuration coming from the config file. I don't think these will need to be different per channel.

rcrichton (commented):
Looking good on a first pass review. I've added some comments.

rcrichton (dismissed):
Great work, this looks awesome.

src/kafkaProducerManager.js
```diff
@@ -0,0 +1,36 @@
+import {KafkaProducer} from './kafkaProducer.js'
+
+export class KafkaProducerManager {
+  static kafkaSet = {}
+
+  static async getProducer(channelName, clientId, timeout) {
+    const kafkaInstance = this.findOrAddConnection(channelName, clientId, timeout)
+    if (!kafkaInstance.isConnected) await kafkaInstance.connect()
+    if (!kafkaInstance.isConnected) throw new Error('Kafka Producer failed to connect.')
```
marrouchi, line 9:
```suggestion
    if (!kafkaInstance.isConnected) await kafkaInstance.connect()
```
await kafkaInstance.connect() should throw the error instead

src/kafkaProducerManager.js
```diff
@@ -0,0 +1,36 @@
...
+
+    return kafkaInstance.producer
+  }
+
+  static findOrAddConnection(channelName, clientId, timeout) {
+    let kafkaInstance = this.getKafkaInstance(channelName, clientId, timeout)
+    if (!kafkaInstance) {
+      kafkaInstance = new KafkaProducer(clientId, timeout)
+      this.kafkaSet[`${channelName}${clientId}${timeout}`] = kafkaInstance
```
marrouchi, line 18: How about we add a urn delimiter to make easy to parse in case we need to : 
```suggestion
      this.kafkaSet[`urn:${channelName}:${clientId}:${timeout}`] = kafkaInstance
```

src/model/channels.js
```diff
@@ -215,6 +223,62 @@ export {RouteDef}
...
+          }
+        }
+      }
+    }
+  ])
+
+  // We need to cross reference the original, not-yet modified, routes
+  // against the incoming dirty routes to see if any were removed and if so remove them from the manager
+  if (Array.isArray(originalKafkaDetails)) originalKafkaDetails = originalKafkaDetails[0]
+  if (originalKafkaDetails && originalKafkaDetails.routes.length > 0) {
```
marrouchi, line 250: is this an object or array ? could we be strict in typing ?

marrouchi (changes requested):

src/model/channels.js
nour-borgi: done :)

src/kafkaProducerManager.js
```diff
@@ -0,0 +1,39 @@
...
+
+    if (kafkaInstance) {
+      if (kafkaInstance.isConnected) await kafkaInstance.disconnect()
+      delete this.kafkaSet[`urn:${channelName}:${clientId}:${timeout}`]
+    }
+  }
+
+  static getKafkaInstance(channelName, clientId, timeout) {
+    return this.kafkaSet[`urn:${channelName}:${clientId}:${timeout}`]
+  }
```
marrouchi, line 38: Let's do this.kafkaSet = new Map();
```suggestion
  static getKafkaInstance(channelName, clientId, timeout) {
    return this.kafkaSet.get(buildURN((channelName, clientId, timeout)))
  }
  
   static buildURN(channelName, clientId, timeout) {
    return return `urn:${channelName}:${clientId}:${timeout}`
  }
```

marrouchi (changes requested):

marrouchi (approved):

PR #1191, opened 2023-04-18
fix: update old password hash (deprecated token strategy)

marrouchi:
## Motivation

Updating the password does not update the old hash value in passport collection. 

How to reproduce  :
- Open OpenHIM (fresh install)
- First login with root@openhim.org / openhim-password
- Reset password (BUG : passwordHash stays the same)
- Run the openhim mapping mediator -> fails to authenticate.

### DIFF
--- a/src/model/users.js
+++ b/src/model/users.js
@@ -107,19 +107,21 @@ export const updateUser = async function (newUserData) {
     if (userToBeUpdated.password) {
       // Compute passport new hash for local strategy
       password = await hashPassword(userToBeUpdated.password)
+
       /* --- @deprecated : Update password hash for token strategy --- */
       const passport = await PassportModelAPI.findOne({
         protocol: 'token',
-        user: user.id
+        email: userToBeUpdated.email
       })
       if (passport) {
-        let shasum = crypto.createHash('sha512');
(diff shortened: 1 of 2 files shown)


src/model/users.js
```diff
@@ -104,8 +105,23 @@ export const updateUser = async function (newUserData) {
...
+      // Compute passport new hash for local strategy
       password = await hashPassword(userToBeUpdated.password)
 
+      /* --- @deprecated : Update password hash for token strategy --- */
+      const passport = await PassportModelAPI.findOne({
+        protocol: 'token',
+        email: userToBeUpdated.email
+      })
+      if (passport) {
+        let shasum = crypto.createHash(passport.passwordAlgorithm)
```
michaelloosen, line 117: Are we sure that this passport.passwordAlgorithm will always have a value? I'm wondering if we shouldn't use this if there is a value and default to a Constant value which can be used in all the places where it needs to be used (instead of hardcoding sha512)

nour-borgi: It was used like this 
image
 
So I will updated to be the same

nour-borgi: done I addressed this with '||'

michaelloosen (approved):

PR #1193, opened 2023-05-18
fix: trust proxy ssl to forward session cookie

marrouchi:
## Motivation

I was deploying OpenHIM for Indonesia on Google Cloud Run. OpenHIM is served behing a proxy that issues the SSL.
It was impossible to connect through OpenHIM console. When checking the logs I found the following error : "Error: Cannot send secure cookie over unencrypted connection"
Basically, you cannot set "secure=true" for the cookie and then serve it in HTTP. For that we needed that option to trust that the proxy.

Check : https://github.com/koajs/koa/issues/974

NOTE : This should be tested in platform.

### DIFF
--- a/config/config.md
+++ b/config/config.md
@@ -46,6 +46,8 @@ The following config option are provided by the OpenHIM. All of these options ha
     "sessionKey": "r8q,+&1LM3)CD*zAGpx1xm{NeQhc;#",
     // If OpenHIM is behind a proxy (should be `true` if the ssl is handled by the proxy)
     "trustProxy": false,
+    // Secure the cookie (either protocol is https or trusting a secured proxy)
+    secureCookie: true,
     // The session max age is the session cookie expiration time (in milliseconds)
     "maxAge": 7200000,
     // The number of characters that will be used to generate a random salt for the encryption of passwords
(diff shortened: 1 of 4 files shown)


rcrichton (dismissed):
LGTM

src/koaApi.js
```diff
@@ -40,12 +40,17 @@ export function setupApp(done) {
 
   // Configure Sessions Middleware
   app.keys = [config.api.sessionKey]
+  
+  if (config.api.trustProxy) {
+    app.proxy = true;
```
arran-standish, line 45:
```suggestion
    app.proxy = true
```
I would just say to remove the terminating ; as the convention is to not user line terminators in openhim-core

config/config.md
```diff
@@ -44,6 +44,10 @@ The following config option are provided by the OpenHIM. All of these options ha
     // The session secret key used for the hashing of signed cookie (used to detect if the client modified the cookie)
     // Signed cookie is another cookie of the same name with the .sig suffix appended
     "sessionKey": "r8q,+&1LM3)CD*zAGpx1xm{NeQhc;#",
+    // If OpenHIM is behind a proxy (should be `true` if the ssl is handled by the proxy)
```
arran-standish, line 47:
```suggestion
    // If OpenHIM is behind a proxy (should be `true` if the proxy sends relevant Forwarded headers)
```
Just updated the wording a bit since it is not related solely to ssl but more being behind a proxy in general

arran-standish (changes requested):
Looks good, just 2 small changes

arran-standish (dismissed):

michaelloosen (approved):

PR #1197, opened 2023-06-27
Return primary response and request body fix

nour-borgi:
Before: the response body of the primary route only will be sent to the next route. 
Now: returning the response and request body of the primary route to the next route.

## Summary by CodeRabbit

- **New Features**
	- Enhanced request and response logging capabilities within the middleware.
	- Introduced new properties for better tracking: `resDetails` for response details and `reqBody` for request data.

- **Bug Fixes**
	- Improved handling of Buffer types for request and response bodies, ensuring accurate logging and processing.


nour-borgi:
Hello Ryan, this PR to replace Hapi proxy by the Kafka channel in eth-cdr (we need the request body and not the response body)

src/middleware/router.js
```diff
@@ -657,7 +667,8 @@ function sendKafkaRequest(ctx, route) {
           method: ctx.request.method,
           path: ctx.request.url,
           headers: ctx.request.headers,
-          body: ctx.body && ctx.body.toString()
+          requestBody: ctx.reqBody ?? '',
+          responseBody: ctx.resBody ?? ''
```
rcrichton, line 671: I'm wondering what else we might need in the future from the response. Perhaps we need to send a structure that similar to what our orchestration's capture (https://github.com/jembi/openhim-core-js/pull/1197/files#diff-65ff1233d04bd7dce50449a41b15b0e5f5dc5226fc06c08dd9fbde9305961c02R482)

This would also introduce a breaking change by getting rid of the `body` property. Perhaps we should keep that to keep backward compatibility.

src/middleware/router.js
```diff
@@ -488,6 +493,11 @@ function sendRequest(ctx, route, options) {
       }
     }
 
+    if (route.primary) {
+      // Save request body of the primary route
+      ctx.reqBody = Buffer.isBuffer(ctx.body) ? ctx.body.toString() : ctx.body
```
rcrichton, line 498: Wouldn't this only be set if the Kafka route is a secondary route. What happen if the Kafka route is a primary route?

rcrichton (changes requested):

rcrichton (approved):

drizzentic (approved):

PR #1200, opened 2023-08-03
Prometheus exporter for OpenHIM

rcrichton:
Creates a prometheus exporter endpoint in the OpenHIM to allow metrics to be scraped by Prometheus. The PR also includes some default OpenHIM dashboards that can be used in grafana to visualise the prometheus metrics.


nour-borgi (approved):
LGTM

PR #1208, opened 2023-09-13
TB-173 Handle transactions stuck in processing state

arran-standish:
If OpenHIM crashes with a request in-flight (so before the down stream service responds) this transaction will stay in 'Processing' forever and does not provide a true reflection of the transaction state.

So on startup it will find all transactions in that state and update them to failed. It uses mongoose's cursor functionality which returns a stream and so shouldn't interfere with the main application as it will only process once the stream receives data. It also properly respects node's event loop and so won't block it (i.e.: incoming requests still get processed normally).


MatthewErispe (approved):
@arran-standish LGTM

rcrichton (approved):
LGTM

PR #1211, opened 2023-10-19
Change Icon type to String and add Application type field to in AppSchema

medchedli:
Replaced the application Icon type from Buffer to String and added application type (which can holds either "link" or "embedded" as values)

### DIFF
--- a/src/model/apps.js
+++ b/src/model/apps.js
@@ -13,6 +13,10 @@ const AppSchema = new Schema({
   description: String,
   icon: {
     type: String  },
+  type: {
+    type: String,
+    enum: ['link', 'embedded'],
+    },
   category: String,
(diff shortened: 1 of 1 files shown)


src/model/apps.js
```diff
@@ -30,4 +32,4 @@ const AppSchema = new Schema({
 })
 
 export const AppModelAPI = connectionAPI.model('App', AppSchema)
-export const AppModel = connectionDefault.model('App', AppSchema)
+export const AppModel = connectionDefault.model('App', AppSchema)
```
nour-borgi, line 35: empty line at the end. Actually you need to run this:  `npm run lint:fix`
It will prettier and lint the code

medchedli:
#1212

PR #1230, opened 2024-06-27
Rework roles and permissions

bradsawadye:
This is a rework of role based access control in the OpenHIM. This extends the basic RBAC we had by adding more permissions. It makes the openhim more flexible and manageable by multiple users in a secure way

## Summary by CodeRabbit

- **New Features**
  - Introduced role-based permission checks for user actions across multiple APIs to enhance security and access control.
  - Added functionality to handle roles for channel access based on permissions.
  - Implemented a new utility function for checking user permissions, improving the flexibility of access control.

- **Refactor**
  - Centralized authorization logic to use a unified permission check function across various API endpoints, improving maintainability.
  - Enhanced transaction management API with refined user authorization checks and a new function to filter sensitive transaction details.

- **Tests**
  - Added new test cases for permission checks on tasks, apps, channels, and roles, ensuring protected operations are only accessible to authorized users.
  - Updated integration tests to validate role-based access control for mediators and metrics, ensuring correct permission handling.


src/api/channels.js
```diff
@@ -132,24 +132,18 @@ export function isMaxBodyDaysValid(channel) {
...
-      ctx,
-      403,
-      `User ${ctx.authenticated.email} is not an admin, API access to addChannel denied.`,
-      'info'
-    )
-    return
-  }
+  try {
+    // Test if the user is authorised
+    const authorised = await utils.checkUserPermission(ctx, 'addChannel', 'channel-manage-all')
```
rcrichton, line 137: The add action should also be controlled by `channel-manage-specified`

src/utils.js
```diff
@@ -228,3 +229,35 @@ export async function hashPassword(password) {
     return Promise.reject(new Error("Password wasn't provided"))
   }
 }
+
+export const checkUserPermission = async (ctx, operation, permission, permissionSpecified, resource) => {
```
rcrichton, line 233: Maybe we could find a way to move the *-specified check into this as well as the check are all very similar in the API endpoints at the moment. It would help to centralise that.

rcrichton (changes requested):
Looking pretty good so far. I've left some guiding comments.

src/utils.js
bradsawadye: it already does. The permissionSpecified is '*-specified' . Example usage of this function - `checkUserPermission(ctx, 'removeChannel', 'channel-manage-all', 'channel-manage-specified', channelId)`

src/api/channels.js
bradsawadye: We cant control this by `channel-manage-specified` because the channel 's `Id` is created when a channel is created on mongo

rcrichton: Ah, I see. Of course.

src/utils.js
rcrichton: Ok, I understand now.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "openhim-core",
   "description": "The OpenHIM core application that provides logging and routing of http requests",
-  "version": "8.4.2",
+  "version": "8.5.0",
```
rcrichton, line 4: I think it might actually be a breaking change as the old way permissions were set for users won't work anymore.

src/api/logs.js
```diff
@@ -15,49 +14,48 @@ const levels = {
...
+
+    ctx.body = results
+    ctx.status = 200
+  } catch (err) {
     utils.logAndSetResponse(
       ctx,
-      403,
-      `User ${ctx.authenticated.email} is not an admin, API access to getLogs denied.`,
-      'info'
+      400,
```
rcrichton, line 56: Did you want to change this to a 400 rather than a 403?

src/api/mediators.js
```diff
@@ -504,21 +487,14 @@ if (process.env.NODE_ENV === 'test') {
...
-      'info'
-    )
-    return
-  }
-
   urn = unescape(urn)
   const config = ctx.request.body
 
   try {
+    const authorised = await utils.checkUserPermission(ctx, 'setConfig', 'mediator-manage-all', 'mediator-manager-specified', urn)
```
rcrichton, line 494: Shouldn't this be:

```suggestion
    const authorised = await utils.checkUserPermission(ctx, 'setConfig', 'mediator-manage-all', 'mediator-manage-specified', urn)
```

src/api/mediators.js
```diff
@@ -564,21 +540,15 @@ function saveDefaultChannelConfig(channels, authenticated) {
...
-      'info'
-    )
-    return
-  }
-
   urn = unescape(urn)
-  const channels = ctx.request.body
 
   try {
+    const authorised = await utils.checkUserPermission(ctx, 'loadDefaultChannel', 'mediator-manage-all', 'mediator-manager-specified', urn)
```
rcrichton, line 546: Shouldn't this be:

```suggestion
    const authorised = await utils.checkUserPermission(ctx, 'loadDefaultChannel', 'mediator-manage-all', 'mediator-manage-specified', urn)
```

rcrichton (changes requested):
Looking good to me, a few more comments.

rcrichton:
@drizzentic @bradsawadye  I just thought of some things while reviewing this:

* We also need a mechanism to migrate people from the old roles to the new ones. Which means a new upgradeDB function (see https://github.com/jembi/openhim-core-js/blob/master/src/upgradeDB.js). This is maybe a separate task and PR to this one. 
* We also need to think if there are any places in the UI where no being able to view something requires changes there. The most obvious would be removing links from the header if they don't have access to manager or view those.

src/api/logs.js
bradsawadye: The permissions check function does not throw the error, it sets the ctx.body and status, and returns. The `catch` here will catch other errors. Will change this to 500

rcrichton (changes requested):
Everything looks good, except I think there are a few permissions that we should remove for simplicity and because certain feature won't work without some of these permissions allows as well. E.g. the certificate managment has it's own permissiosn yet if the service-manage permissions isn't also set then setting a new certificate won't work.

drizzentic:
> ervice-manage

I have removed the authorization on the restart function.

rcrichton (approved):

